PR #51, opened 2020-08-15
Fix icon logic, closes #50

### LINKED ISSUE #50: Icons in /usr/share/icons/hicolor/256x256/ don't seem to work
If I read the code correctly, then the code looks for the presence of the icon specified in the desktop file in `appdir + /usr/share/icons/hicolor/256x256/` here:
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L369-L382

What strikes me though is that in the first check it is checked whether the icon exists at the root of the appdir. This in and of itself is not what I find interesting. Rather the fact that this code part sets the `iconFile` variable:
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L367-L369

The second check however doesn't seem to set this variable. Thus
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L386
Prints 
```
Icon file: 
```
aka: An empty String.

A few lines down however this variable is used to assemble a path to copy:
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L397

In my case this results in this error:
```
ERROR Copy .DirIcon: read .: is a directory
```

I also think it is interesting that the code above seems to create a `.DirIcon` file in the AppImage's root
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L371-L381

and a bit further down we have
https://github.com/probonopd/go-appimage/blob/f1ebcce70dae0ab9a671604261d4fd46be88384f/src/appimagetool/appimagetool.go#L391-L394
which seems to exactly cancel above action. Interestingly however I did not see the `Deleting pre-existing .DirIcon` message in my logs so maybe I am missing something here :thinking: 

All in all I think that maybe the functionality for icons in the respective subdir is broken atm...

probonopd:
Thanks for your careful review @Krzmbrzl

### DIFF
--- a/src/appimagetool/appimagetool.go
+++ b/src/appimagetool/appimagetool.go
@@ -362,32 +362,29 @@ func GenerateAppImage(appdir string) {
 	var iconfile string
 
 	// Check if we find a png matching the Icon= key in the top-level directory of the AppDir
+	// or at usr/share/icons/hicolor/256x256/apps/ in the AppDir
 	// We insist on a png because otherwise we need to costly convert it to png at integration time
 	// since thumbails need to be in png format
 	if helpers.CheckIfFileExists(appdir+"/"+iconname+".png") == true {
 		iconfile = appdir + "/" + iconname + ".png"
-	} else if helpers.CheckIfFileExists(appdir + "/usr/share/icons/hicolor/256x256/" + iconname + ".png") {
+	} else if helpers.CheckIfFileExists(appdir + "/usr/share/icons/hicolor/256x256/apps/" + iconname + ".png") {
 		// Search in usr/share/icons/hicolor/256x256 and copy from there
-		input, err := ioutil.ReadFile(appdir + "/usr/share/icons/hicolor/256x256/" + iconname + ".png")
+		input, err := ioutil.ReadFile(appdir + "/usr/share/icons/hicolor/256x256/apps/" + iconname + ".png")
 		if err != nil {
 			log.Println(err)
 			return
 		}
-		err = ioutil.WriteFile(appdir+".DirIcon", input, 0644)
-		if err != nil {
-			log.Println("Error copying ticon to", appdir+".DirIcon")
-			log.Println(err)
-			return
-		}
+		iconfile = appdir + "/usr/share/icons/hicolor/256x256/apps/" + iconname + ".png"
 	} else {
-		os.Stderr.WriteString("Could not find icon file at " + appdir + "/" + iconname + ".png" + ", exiting\n")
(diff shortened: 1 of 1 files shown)


Krzmbrzl:
You're welcome. Thanks for the fix :D

src/appimagetool/appimagetool.go
```diff
@@ -362,32 +362,29 @@ func GenerateAppImage(appdir string) {
...
+	// or at usr/share/icons/hicolor/256x256/apps/ in the AppDir
 	// We insist on a png because otherwise we need to costly convert it to png at integration time
 	// since thumbails need to be in png format
 	if helpers.CheckIfFileExists(appdir+"/"+iconname+".png") == true {
 		iconfile = appdir + "/" + iconname + ".png"
-	} else if helpers.CheckIfFileExists(appdir + "/usr/share/icons/hicolor/256x256/" + iconname + ".png") {
+	} else if helpers.CheckIfFileExists(appdir + "/usr/share/icons/hicolor/256x256/apps/" + iconname + ".png") {
 		// Search in usr/share/icons/hicolor/256x256 and copy from there
-		input, err := ioutil.ReadFile(appdir + "/usr/share/icons/hicolor/256x256/" + iconname + ".png")
+		input, err := ioutil.ReadFile(appdir + "/usr/share/icons/hicolor/256x256/apps/" + iconname + ".png")
```
Krzmbrzl, line 372: Why do you read the image here but not in the branch above? :thinking:

probonopd: Good question - probably just using `helpers.CheckIfFileExists` would be sufficient...

Krzmbrzl: That could also be outsourced to after the if. Just check whatever is in`fileName`. That way you'd also get a consistent behavior across all branches :)

probonopd: Care to send a PR @Krzmbrzl? I'd highly appreciate it.

Krzmbrzl: I can try - never worked with Go before ^^

probonopd: Looks like you succeeded :+1: Thank you.

PR #55, opened 2020-08-16
Remove unneeded file read

Krzmbrzl:
If the icon was stored in `.../icons/hicolor/256x256/apps`, the code would read the file's content into memory. It doesn't do anything with that though. And the existance of the file is covered by the if-condition anyways.

Thus this commit removes the unnecessary file-access.


probonopd:
Thanks @Krzmbrzl

Krzmbrzl:
Sure :)

PR #78, opened 2020-10-22
feat: Allow providing manual branch name

srevinsaju:
Currently, go-appimage does not allow to provide the branch / tag name / channel from where
github releases can be updated (updateinformation). This is useful for releasing apps to, `beta`, `nightly` channels, by setting APPIMAGETOOL_GH_BRANCH_DEPLOY environment variable to the required branch/channel name.

Use case, in the [Firefox-AppImage](/srevinsaju/Firefox-AppImage), I would like to maintain Beta, Nightly, Stable, ESR Releases in a single repository. I could achieve this by doing something like this

```bash
export APPIMAGETOOL_GH_BRANCH_DEPLOY=nightly
./appimagetool*.AppImage AppDir
```

I am open to changes / feel free to close this PR if it seems redundant

### DIFF
--- a/src/appimagetool/appimagetool.go
+++ b/src/appimagetool/appimagetool.go
@@ -555,7 +555,9 @@ func GenerateAppImage(appdir string) {
 		} else {
 			parts := strings.Split(os.Getenv("GITHUB_REPOSITORY"), "/")
 			var channel string
-			if os.Getenv("GITHUB_REF") != "" && os.Getenv("GITHUB_REF") != "refs/heads/master" {
+			if os.Getenv("APPIMAGETOOL_GH_BRANCH_DEPLOY") != "" {
+				channel = os.Getenv("APPIMAGETOOL_GH_BRANCH_DEPLOY")
+			} else if os.Getenv("GITHUB_REF") != "" && os.Getenv("GITHUB_REF") != "refs/heads/master" {
 				channel = "latest"
 			} else {
 				channel = "continuous"


srevinsaju:
@probonopd kindly provide your suggestions :innocent:

srevinsaju:
I guess, it would be better to rename the environmental name
```bash
APPIMAGETOOL_GH_BRANCH_DEPLOY -> APPIMAGETOOL_GH_CHANNEL_DEPLOY
```
to make it more meaningful

probonopd:
Thanks @srevinsaju. Were you able to test this fully, including that updates work when using multiple different channels?

srevinsaju:
I tested it [here](https://github.com/srevinsaju/go-appimageteststeststestestests), sorry for the repository name :smile: 

* After setting the custom branch: https://github.com/srevinsaju/go-appimageteststeststestestests/actions/runs/322066167
* Before setting the custom branch: https://github.com/srevinsaju/go-appimageteststeststestestests/actions/runs/322061699

probonopd:
That looks good. Could you perform an end-to-end test, actually updating, when you have more than one channel?

srevinsaju:
I did not understand, did you mean to implement this in Firefox-AppImage or a similar repository, so that you will be able to see?

probonopd:
Yes, if that's not too much to ask :-)

srevinsaju:
Ok, may I know if I would be able to download the build artifact from Travis CI? Because, it would make things easier for me  :smile: , if not, I will try to build it on an old system

srevinsaju:
Ok, I am porting the scripts/build.sh to GitHub actions, I am not sure if you are its fan, but I will try it on my fork :smile:

probonopd:
GitHub Actions is something that I'd like to support anyway. When I had started working on go-appimage it was not there yet (or at least I had no clue about it at that time).

srevinsaju:
Sure, I have created a go-appimage GitHub Action to build every commit and provide an artifact for download for PRs with their base branch as `dev` on https://github.com/srevinsaju/go-appimage

Regarding adding it to Firefox-AppImage, because of #68 I will not be adding this yet, but I would try to fix #68 before this, as its quite important. Its the only blocker which I see now, which prevents me from using go-appimagetool. Until then, I will create nightly and beta builds of go-appimagetool on my fork as a demo. Right now, you can see the releases for x86_64 and i386 are generated using your script.

probonopd:
Good. Just ping me when you think everything is end-to-end tested well enough that you consider this safe to merge. Thanks!

srevinsaju:
@probonopd Done!, you can see [this Firefox repository](/srevinsaju/.tmp.Firefox.AppImage) and its channels.

probonopd:
Thank you very muich @srevinsaju for the implementation and testing. Could you please add a sentence on how to use it to the README.md? Thank you very much.

srevinsaju:
@probonopd I understand your idea, but I feel that the core tools should be extensively configurable, and end-users should  never use low level tools like appimagetool, but instead, they should use higher level tools like pkg2appimage, or appimage-builder. Restricting the ability to configure low level applications makes it almost impossible to configure them on higher level helper tools (unless we fork go-appimage)


Regarding branches, actually, the right term I should have used is `channel`. For example, Firefox-AppImage is built on the `master` branch and releases are created as `tags`. We cannot retrieve the tags from gh actions, gitlab or travis because we set them after we build the executable right? Moreover, I would consider this one environment variable as protected variable, i.e. the user should only set this if they 'really know what they're doing™'. I would suggest printing a warning message to stdout, what do you suggest?

probonopd:
The name `appimagetool` in go-appimage is a misnomer: It is not the same low-level tool as the old `appimagetool` in https://github.com/AppImage/AppImageKit. Instead, it is a very high-level tool intended to do everything, including uploading, so that the user has zero chance to mess anything up. If something is not working when using this tool, then the tool itself is to be blamed and needs to be fixed, rather the user trying to do random workarounds that do more harm than good.

Let's come up with a better name for this tool, something that conveys "High-level fully-automagic zero-configuration AppImage all in one creation tool".

srevinsaju:
oh I see. I thought its a new rewrite of the old appimagetool, which was supposed to be the low-level tool which was supposed to just convert an AppDir to appimagetool.

probonopd:
It isn't - it's a naming screwup. (It started out being that, but then I just kept coding, adding more and more functionality toward a full blown "do it all" tool, so the name no longer is a good match.)

srevinsaju:
We could ask for suggestions on a new issue for a cool new name. The community would be interested to provide some inputs, but this repository is less discoverable than [AppImageKit](https://github.com/appimage/appimagekit), would this repository be moved some day?

PR #80, opened 2020-10-22
fix: non-travis builds contain illegal character ':'

srevinsaju:
The %T expanded to 00:00:00 which is contains the character `:`. Many systems including shells, CIs do not like to keep this character and causes errors, or sometimes we need to escape these characters. It would be nice if the character is removed


probonopd:
Good catch @srevinsaju, thank you.

PR #81, opened 2020-10-23
Fix arch detection

srevinsaju:
Previously, go-appimagetool did not stop if a proper architecture was
not detected from the .so. files or the binary. As it did not halt,
the arch variable was set to empty string (''). This led to a
filenotfound panic, because the tool was looking for
`release-` instead of `release-x86_64` as the arch variable was
empty. go-appimagetoold should properly halt if the detection of
ARCH variable did not complete successfully, so that they
can provide the ARCH environment variable instead.

### DIFF
--- a/src/appimagetool/appimagetool.go
+++ b/src/appimagetool/appimagetool.go
@@ -6,8 +6,11 @@ package main
 import (
 	// "crypto/md5"
 	"encoding/json"
+	"errors"
 	"flag"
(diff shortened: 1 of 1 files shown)


srevinsaju:
@probonopd kindly review and provide your suggestions :innocent:

probonopd:
Thank you very much, this is a very reasonable point. Merging.

PR #85, opened 2020-10-24
fix: regression introduced by 71d52986d56ab486f97f4e6a998a1d7495b47e3c

### LINKED ISSUE #84: Build fails
Unfortunately the build now fails.
https://github.com/probonopd/go-appimage/runs/1301568998

Seems to be related to dc8b62c?

cc @srevinsaju

srevinsaju:
Fixes #84 

needed to predefine the variable before the if block
as scopes in go are also limited to the if-else-block only.


probonopd:
Thank you @srevinsaju

PR #91, opened 2020-11-01
Restructure and rewrite 

### LINKED ISSUE #87: Build a bare-bones version of appimagetool
Once https://github.com/probonopd/go-appimage/issues/86 is done, we may want to also build a bare-bones version of the tool that does only one thing, turn AppDirs into AppImages (like the old appimagetool in https://github.com/AppImage/AppImageKit did). This might make it more viable for other projects (e.g., electron-builder, linuxdeployqt, appimage-builder,...) to use it under the hood.

That one might possibly be called `appimagetool` and might be made command-line compatible to the old `appimagetool` from AppImageKit, which at that point might be deprecated.

While I don't have a need for this myself (I will use the do-it-all-automagic full blown tool), I'd welcome PRs that would build that bare-bones version.

Only requirement:
Things don't get more complicated for me. Especially, no multi-repository setups.

cc @TheAssassin @srevinsaju

### LINKED ISSUE #90: Cleaning up and restructuring - Part 1
image

I am restucturing `go-appimagetool` following #87 (comment) to get a better structure. This is the command line interface so far which was accompanied with the restructuring. Let me know of your ideas. comments so that I can incorporate in the PR.

It looks a bit more neater for me, but this is followed by a lot of other changes too. 

cc @probonopd

srevinsaju:
May fix #90 
Fixes #87

In progress.
Rewriting `appimagetool` following `urfave/cli/v2` to write cleaner command line interfaces

cc @probonopd 
open for discussion, and rewrite

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	github.com/srwiley/oksvg v0.0.0-20200311192757-870daf9aa564
 	github.com/srwiley/rasterx v0.0.0-20200120212402-85cb7272f5e9
+	github.com/urfave/cli/v2 v2.2.0
 	go.lsp.dev/uri v0.3.0
 	golang.org/x/crypto v0.0.0-20200311171314-f7b00557c8c4
 	golang.org/x/image v0.0.0-20200119044424-58c23975cae1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,4 @@
+github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d h1:G0m3OIz70MZUWq3EgK3CesDbo8upS2Vm9/P3FtgI+Jk=
 github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/acobaugh/osrelease v0.0.0-20181218015638-a93a0a55a249 h1:fMi9ZZ/it4orHj3xWrM6cLkVFcCbkXQALFUiNtHtCPs=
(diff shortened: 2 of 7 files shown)


probonopd:
Impressive @srevinsaju. Thank you.

> The AppImage was created, but is lacking update information.

Do you know why this might be?

srevinsaju:
I guess its because update information are not calculated because of PULL REQUEST setting in Travis. I tested locally by exporting GITHUB_REPOSITORY, it seems to work on the GitHub actions part. It should work when its merged to master.

probonopd:
That would be an explanation, yes. (I hope it will work this way.)

srevinsaju:
Few questions:
* Would you like it if I moved all the command line interfaces, to a file called `cli.go` or `interface.go` and called those functions from  `appimagetool.go` ? I do not mind keeping it either place, both looks neat.
* Is the current pull request too messy, does it keep the functional flow? Or should I make any changes?

probonopd:
From glancing over it quickly, I wonder whether we can do without passing around `options` so much. What do you think?

srevinsaju:
Yes, I agree, I do not like passing `options`  everywhere.
In python, I would have simply created an object, and then passed the value in the object creation
```python
class Object:
   def __init__(self, options):
      self.options = options

   def makeAppImage(self):
       # do something
```

How can we do that in `go`? I have not yet experimented with Object Oriented programming in `go`, let me research on how to make it better. Let me know if you know any alternatives to writing that part too

srevinsaju:
Okay, I declared a single variable `options` globally. It looks right now!

srevinsaju:
@probonopd  Ready for review!

probonopd:
Hello @srevinsaju. Are you using an IDE? E.g., GoLand has some checks built in. One of the checks determines whether errors are handled and warns if they are not.

Hence, please leave in things like `_ = ...`.

Most of the time _ point explicitly out that we are ignoring errors. Potentially in some (but not all) of those cases we may want to add explicit error handling. Ideally we would state the rationale in a comment in the (rare) cases where we do not want errors to be handled explicitly.

So, please do not remove any `_ = ...` and similar constructs. They are there intentionally It allows one to search for them, and iirc the GoLand IDE  complains if those are not present.

srevinsaju:
@probonopd :smile: yes, I switched from `vim` to `GoLand` because of its cool features. I thought its a convention. I have rebased the PR to remove the explicitly ignoring part

probonopd:
Thanks @srevinsaju. Do you know a better method of generating a random password?

srevinsaju:
@probonopd `crypto/rand` is a good place to get better random numbers. those provided by `math/random` are pseudo-random numbers, whose pattern is known

srevinsaju:
* math/rand implements a large selection of pseudo-random number generators.
* crypto/rand implements a cryptographically secure pseudo-random number generator with a limited interface.

source: https://yourbasic.org/golang/crypto-rand-int/#:~:text=math%2Frand%20implements%20a%20large,generator%20with%20a%20limited%20interface.

probonopd:
@srevinsaju I'd say we tackle the password topic in another PR; shall we?

srevinsaju:
@probonopd yes I agree.

srevinsaju:
:crossed_fingers: lets see if the update information gets embedded as expected theoritically :smile:

probonopd:
Thanks a lot @srevinsaju for your large go-appimage contribution.
For future PRs, please try to make them address one thing per PR.

srevinsaju:
Yes, sure. I tried to, but I did was "Restructure and rewrite", so first I rewrote the go-appimage and then I committed. I did not know what I was going to change in the beginning, I was manually tweaking and playing with the code, and then finally it reached to this. Hopefully, this was the last "restructure and rewrite" PR from me which covered most parts. The rest of them are minor PRs :smile:. I will make the rest of them atomic

PR #92, opened 2020-11-03
Fix finding gst-plugin-scanner

jonaski:


probonopd:
Thanks @jonaski. Which problem is this PR trying to solve (e.g., on which distribution it did not work correctly before)?

jonaski:
It fixes finding and setting gst-plugin-scanner on openSUSE where the executable is called gst-plugin-scanner-x86_64

probonopd:
Hi @jonaski why did you close this PR?

jonaski:
I didn't think you'd accept the PR, also I realize more work was required to make this work.
A wrapper script is required that calls the actual binary so that LD_LINUX is set like you've done in AppRun.
So I've modified the AppRun to set the GST_PLUGIN_SCANNER variable to point to a script called gst-plugin-scanner.sh instead of the actual binary, this script launches the actual gst-plugin-scanner binary.
See: https://github.com/strawberrymusicplayer/strawberry-buildbot/blob/master/config/dist/gst-plugin-scanner.sh
Do want me to open a new PR that includes this?

Strawberry also runs another binary called strawberry-tagreader internally to identify and scan tags in audio files, without it it won't recognize any local audio files, I did a similar thing for that one. After 'make install' I simply rename "strawberry-tagreader" "strawberry-tagreader-bin", then copy in a script in place with the same name (strawberry-tagreader), and it will call "strawberry-tagreader-bin". It now works on older distros too, before it failed launch strawberry-tagreader which means it wasn't able to load any local files. This is of course unique for strawberry, but maybe some example template could be included for programs that does similar things.
You can see the whole process on MakeAppImageBuilder in https://github.com/strawberrymusicplayer/strawberry-buildbot/blob/master/config/master/strawberry/builders.py
I deleted the previous builds from https://builds.strawberrymusicplayer.org/appimage/ so all the files there are working with tagreader and plugin scanner.

go-appimage seem to be the best tool at the moment, but I have to heavily modify it for everything to work.

probonopd:
> So I've modified the AppRun to set the GST_PLUGIN_SCANNER variable to point to a script called gst-plugin-scanner.sh 

Could this be done within the AppRun shell script without the need for an extra file? We should also document in the code why we are doing this.

PR #93, opened 2020-11-03
Add glib-networking module

jonaski:
The glib-networking module (libgiognutls) is required to stream HTTPS with gstreamer.
Setting GIO_EXTRA_MODULES makes GStreamer find it.
The search path here should probably not be fixed here though.

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -94,7 +94,8 @@ if [ ! -z $(find "${HERE}" -name "libgstcoreelements.so" -type f) ] ; then
   export GST_PLUGIN_PATH=$(dirname $(readlink -f $(find "${HERE}" -name "libgstcoreelements.so" -type f | head -n 1)))
   export GST_PLUGIN_SCANNER=$(find "${HERE}" -name "gst-plugin-scanner" -type f | head -n 1)
   export GST_PLUGIN_SYSTEM_PATH=$GST_PLUGIN_PATH
-  env | grep GST
+  export GIO_EXTRA_MODULES=$(dirname $(readlink -f $(find "${HERE}" -name "libgiognutls.so" -type f | head -n 1)))
+  env | grep "GST_\|GIO_"
 fi
 
 ############################################################################################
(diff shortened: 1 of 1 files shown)


jonaski:
I can adapt the similar approach you use for gst-plugin-scanner, or we can search the entire /usr.

src/appimagetool/appdirtool.go
```diff
@@ -665,6 +666,7 @@ func handleGStreamer(appdir helpers.AppDir) {
 			} else {
 				log.Println("Bundling dependencies of GStreamer 1.0 directory...")
 				determineELFsInDirTree(appdir, locs[0])
+				determineELFsInDirTree(appdir, "/usr/lib64/gio/modules/libgiognutls.so")
```
probonopd, line 669: I fear that the `/usr/lib64/` is specific to your distribution. Fedora-type distributions tend to use `/usr/lib64/` whereas others tend to use `lib/x86_64-linux-gnu`...

probonopd (changes requested):
I fear that the `/usr/lib64/` is specific to your distribution. Is there a way to not hardcode it?

src/appimagetool/appdirtool.go
jonaski: I agree, like I already wrote I would like your input on how you want to do it. I'm not a Go programmer so I'm learning as I work on this.
The only distro I've only seen use /usr/lib/x86_64-linux-gnu is debian and derivatives of debian (which most are basically debian or ubuntu variants as most debian distros use the same packaging).
Everyone else use /usr/lib or /usr/lib64, this is common for Redhat/Fedora/Centos, SUSE/openSUSE, Mageia, Arch and most other I've seen.

probonopd: > The only distro I've seen use /usr/lib/x86_64-linux-gnu is debian and derivatives of debian

...which makes the vast majority in terms of market share of Linux distributions.

probonopd:
Hi @jonaski why did you close this PR?

jonaski:
I didn't think it was going anywhere, it turned into a Linux distro discussion.

probonopd:
Sure I would happily merge it if it could be written in a way that does not make assumptions about paths (hardcoded). I have avoided hardcoding paths in the rest of the code as you may have seen.

PR #94, opened 2020-11-13
Update README.md

mihaiav:
I think it should be AMD64 not ARM64. ARM64 is not x86_64


probonopd:
Great catch, thank you.

PR #96, opened 2020-11-18
Quick and dirty patch to stop uca.xml overwriting

### LINKED ISSUE #95: Thunar custom actions file gets overwritten by appimaged on startup
Thunar's custom actions file (~/.config/Thunar/uca.xml) gets overwritten by appimaged. This is still happening as of appimaged 582 and the first version I installed on this computer (552, I believe) also had the issue.

After appimaged starts up, the uca.xml file is overwritten by one that only contains appimaged's "Update - Update the appimage" action, and all my custom actions such as "Open terminal here", etc are lost.

I'm running appimaged 582 and Thunar 4.15.3 on Gentoo. I'll be happy to provide any other details necessary! :)

IvanDSM:
This commit adds a patch to stop Thunar's uca.xml file being overwritten on XFCE. Fixes #95 

This patch introduces a check to see if our action is present in the user's uca.xml, and in case it isn't, appends it to the beginning of the <actions> section of uca.xml.

I apologize if the quality of the code isn't good, this is my first time writing Go and I'm mainly a C++ programmer :(

### DIFF
--- a/src/appimaged/filemanager.go
+++ b/src/appimaged/filemanager.go
@@ -70,7 +70,7 @@ Name=Make executable
     <unique-id>` + XFCEThunarActionUniqueID + `</unique-id>
     <command>` + arg0abs + ` %f</command>
     <description>Update the AppImage</description>
-    <patterns>*AppImage,*.appimage</patterns>
+    <patterns>*.AppImage;*.appimage</patterns>
     <other-files/>
     <directories/>
 </action>
--- a/src/appimaged/filemanager.go
+++ b/src/appimaged/filemanager.go
@@ -62,10 +62,12 @@ Icon=utilities-terminal
 Name=Make executable
 `
 
+	XFCEThunarActionUniqueID := `1573903056061608-1`
+
 	XFCEThunarAction := `<action>
     <icon>terminal</icon>
     <name>Update</name>
-    <unique-id>1573903056061608-1</unique-id>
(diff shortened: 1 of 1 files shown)


probonopd:
Thank you very much @IvanDSM for your PR. Highly appreciated!

I think we should turn "1573903056061608-1" into a variable that is used in all places, to ensure that the value is not inadvertently changed in one but not the other location(s).
Could you please make this change?

Another thought, do the actions work 1:1 or do they need to be rewritten as well? I could imagine that we need to use `arg0abs` in the `Exec=` line for the actions as well?

How did you test the success of your patch, and does it work satisfactorily for you as-is?

Again, thanks a lot!

IvanDSM:
Hello there! I will do the extraction of the unique id to a variable :)

Regarding how I tested it, I made a uca.xml without the "Update the Appimage" action and tested to see if my patched version of go-appimaged would add the action to the file without overwriting. This succeeded, and next I tried restarting my patched version to see if it would correctly identify the action as being there and not add it again to the file. This also succeeded.

I'm not quite sure what you mean with the arg0abs observation :( As for the action, I did find issues with the action as-is that prevented it from appearing, and I will also fix those in a commit.

probonopd:
Thanks @IvanDSM. We will have to use `arg0abs` when we want to make actions of other applciations (not  "Update the Appimage") to work, I guess.

Issue #76, opened 2020-10-20
Irregularities in appimagetool.go and appdirtool.go


Juuliuus:
Hi. As I mentioned at [https://discourse.appimage.org/t/is-it-possible-with-s-deploy-to-include-libraries-loaded-at-runtime/1779/40](url) I had found some irregularities while working with appimagetool.go and appdirtool.go, just things I noticed, maybe they are, or maybe are not, problems.

I don't really know if you want separate issues for each of these? So I will list them here and just let me know what to do, and I will do it. 

1. in both "-s deploy" and in "-s -l deploy" (the l switch is no longer needed I think) there are two copies of ld-linux-x86-64.so.2 in different folders(/lib64/ and in /usr/lib64/, and similar in -l deploys). The sha256 sums of  both are different. In my testing I always used the /lib64/ version and it worked.

2. In the RPath of the patched libraries one finds "$ORIGIN/gconv" which is never correct, on my builds it should be "$ORIGIN/lib/x86_64-linux-gnu/gconv". The gconv path is export'd to environment correctly, however.

3. RPaths also have an "$ORIGIN/..", doesn't hurt but doesn't help? 

4. Running "-s deploy" a second time fails in the qt5 sections, I don't know if appimagetool expects a clean, "fresh" appdir or not, but once deployed one cannot deploy again. Perhaps this is by design?

probonopd:
4 is by design, the others are not. Thanks for reporting. Unfortunately right now I don't have the time to look into them, maybe you can have a look at the code?

Juuliuus:
Happy to.

Juuliuus:
Sorry I thought I had posted that I'd be happy to, but never hit comment! Anyway, here are the solutions.

== deployElf ==

The 2nd ld-linux comes from deployElf. In the slice allELFs there is a "/usr/lib64/ld-linux-x86-64.so.2" string that it found
earlier. Apparently analysis of some included libraries referred to that file and so it was added to allELFs. Since the deployInterpreter function correctly handles ld-linux from the /lib64/ folder any other references to ld-linux need to be ignored. There may be subtleties on some systems such that the routine(s) that build allELFs needs to be tightened up.

So the easy fix is to check for the ld- prefix in deployElf and ignore that/those entries. 

> if strings.HasPrefix(filepath.Base(lib), "ld-") == true {
> 	log.Println("deployElf: Ignore ", lib, "because it is already deployed...")
> 	return
> }
> 

[...]

PR #97, opened 2020-11-20
Stop superfluous copy of ld-linux so

Juuliuus:
deployELF function was copying ld-linux so from locations other than /lib64/. Use a similar routine in deployELF as is used in patchRpathsInElf to prohibit ld-linux so processing. #76

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -412,7 +412,12 @@ func deployElf(lib string, appdir helpers.AppDir, err error) {
 			return
 		}
 	}
-	
+
+	if strings.HasPrefix(filepath.Base(lib), "ld-") == true {
+		log.Println("deployElf: Ignore", lib, "because it is already deployed...")
+		return
+	}
+
 	log.Println("Working on", lib, "(TODO: Remove this message)")
 	if strings.HasPrefix(lib, appdir.Path) == false { // Do not copy if it is already in the AppDir
 		libTargetPath := appdir.Path + "/" + lib


probonopd:
Are you sure it's not needed in _both_ locations (e.g., because it was in _both_ locations on the original system)?

Juuliuus:
No, I'm not sure. What I do know: 

This fix was used in all my testing and the resulting appimage worked flawlessly. 

At the time I noticed the problem I tracked down the "extra" ld-linux. 
It had copied /usr/lib64/ld-linux-x86-64.so.2  to appdir/usr/lib64 (this is the superfluous copy).

There were two other d-linux's as well.. /lib64/ld-linux-x86-64.so.2 (the one copied and patched by appimagetool), and /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2.

All 3 of these ld-linux's are links to the same 187.0 KiB file: /lib/x86_64-linux-gnu/ld-2.31.so

In addition appimagetool writes this line (which comes out of patchRpathsInElf) right after the ld-linux file is/would be copied:
Not writing rpath in /home/juus/RCWork/AppImageDEV/libappimage_hooks_CDoo/commandoo.AppDir/usr/lib64/ld-linux-x86-64.so.2 because its name starts with ld-...

And the last reference to the ld-linux file is this line:
Relative path from ld-linux to Qt prefix directory in the AppDir: ../usr/lib/x86_64-linux-gnu/qt5
Which is correct for appdir/lib64, but not appdir/usr/lib64

And think about this: At this point in the code the ld-linux file has already been copied and patched. Why would another (unpatched) copy somewhere else be of use?

I am not as knowledgeable about the linux file jungle as you are. Does it have knock on effects? I don't know, I've only seen this on my system, and in this case the "superfluous" copy is a direct copy of the same ld-linux in appdir/lib64, just not patched. I'm assuming this /usr/... copy got into deployELF because it was referenced by some other library being processed as a dependency. But we all know how assumptions work out...

So, bottom line: I guess I can not promise that it will not break something. Maybe in that case just leave it alone?

probonopd:
Well, the truth is I am not working with Linux much anymore these days. So I never came around to testing this systematically. If someone has some time, systematic testing of this topic would be highly appreciated. Thanks!

PR #103, opened 2020-11-24
feat: get that candy progressbar which appimagetool has when it builds appimages

srevinsaju:
@probonopd :smile: :innocent:


src/appimagetool/appimagetool.go
```diff
@@ -778,95 +778,3 @@ func GenerateAppImage(appdir string) {
 	fmt.Println("at https://github.com/AppImage/appimage.github.io")
 }
 
-
-// main Command Line Entrypoint. Defines the command line structure
```
probonopd, line 782: Is it correct that this PR wants to remove this whole section?

srevinsaju: uh oh; this commit has been included in #104; shall I close this?; looks like my mistake

srevinsaju: The net change, I wanted was L576 smh; maybe I should rewrite this

PR #104, opened 2020-11-24
Core AppImage Tool 

### LINKED ISSUE #99: Split out a low-level tool like the old appimagetool
Split out a low-level tool like the old appimagetool

srevinsaju:
Fixes #99 
WIP

https://github.com/probonopd/go-appimage/pull/104/commits/64e51f6dcb71856c0369d97a69431bdb233a5e6e should be related to #103


probonopd:
Hi @srevinsaju, thanks for working on this. Did you notice the last couple of builds failed with

```
can't load package: package github.com/probonopd/go-appimage/src/mkappimage: cannot find package "github.com/probonopd/go-appimage/src/mkappimage" in any of:
	/usr/local/go/src/github.com/probonopd/go-appimage/src/mkappimage (from $GOROOT)
	/tmp/go/src/github.com/probonopd/go-appimage/src/mkappimage (from $GOPATH)
```

srevinsaju:
Yes, I did notice; but I do not know why; I added the travis build script on https://github.com/probonopd/go-appimage/pull/104/commits/cc3b230a1d54789ebdbff26182e1e4f0095790a9, I hope I got that one right;

If I guess right, maybe its because the go api checks on `master` branch, and finds no `mkappimage` there? I do not know how go manages packages tho :cry:, I would need your help

srevinsaju:
Will fix #103 shortly :rocket:

srevinsaju:
Should be fixed by merging #109 , @probonopd kindly review

probonopd:
Where would this upload the build artefacts?

https://github.com/srevinsaju/go-appimage/actions/runs/387706699 is not ideal because those are zips. We need the "naked" AppImages on GitHub Releases.

You can use e.g., the https://github.com/tcnksm/ghr tool to accomplish that.

srevinsaju:
To github releases, using marvinpinto/automatic-release-action, it will only be released, if the target branch is `master`, and not a pull request

probonopd:
Excellent. Will they be unzipped?

srevinsaju:
Yes, example here: https://github.com/srevinsaju/guiscrcpy/releases/latest

I use the same action, so it would work exactly the same.
That zip from the artifacts will be unzipped, and each item from the zip will be released as one asset each

scripts/build.sh
```diff
@@ -214,3 +259,30 @@ Terminal=true
...
+chmod +x mkappimage.AppDir/usr/bin/*
+cp mkappimage-$USEARCH mkappimage.AppDir/usr/bin/mkappimage
+( cd mkappimage.AppDir/ ; ln -s usr/bin/mkappimage AppRun)
+cp $GOPATH/src/github.com/probonopd/go-appimage/data/appimage.png mkappimage.AppDir/
+cat > mkappimage.AppDir/mkappimage.desktop <<\EOF
+[Desktop Entry]
+Type=Application
+Name=mkappimage
+Exec=mkappimage
+Comment=Optional daemon that integrates AppImages into the system
```
probonopd, line 282: Is this correct? Probably a copy&paste leftover?

piegamesde:
I'm trying this out (building from source), and it says `2021/05/28 23:37:59 Required helper tooluploadtoolmissing`. I could not find anything under that name (I also looked for `uploadtool` in the assumption that the formatting is broken).

internal/helpers/helpers.go
```diff
@@ -332,6 +332,18 @@ func CheckIfSquashfsVersionSufficient(toolname string) bool {
...
 }
 
+// CheckIfAllToolsArePresent checks if all the required tools, for example. mksquashfs, unsquashfs
+// are present using exec.LookPath. Function exits with exit code 1, if any tool is missing
+// and reports that particular missing tool to the user
+func CheckIfAllToolsArePresent(tools []string) {
+	for _, t := range tools {
+		_, err := exec.LookPath(t)
+		if err != nil {
+			log.Fatal("Required helper tool", t, "missing")
```
piegamesde, line 342:
```suggestion
			log.Fatal("Required helper tool '", t, "' missing")
```

piegamesde:
So I managed to work around all the rough edges so far (I can't tell if they are caused by this PR or not, so I'll open issues once this is merged), until I ran into this error:

```
mksquashfs AppDir foo-bar-x86_64.AppImage -offset 189632 -fstime 1622247904 -comp gzip -root-owned -noappend
FATAL ERROR:SOURCE_DATE_EPOCH and command line options can't be used at the same time to set timestamp(s)
ERROR mksquashfs: exit status 1
```

srevinsaju:
I tested this branch. I was able to create a working appimage. Looks like either of this happened:
* you are using an older version of `mksquashfs`
* You have set `SOURCE_DATE_EPOCH` environment variable.

src/mkappimage/mkappimage.go
```diff
@@ -0,0 +1,232 @@
...
+// provided as arguments exists. If yes add the current path to PATH,
+// check if all the necessary dependencies exist,
+// finally check if the provided argument, AppDir is a directly.
+// Call GenerateAppImage with the converted arguments
+// 		Args: c: cli.Context
+func bootstrapMkAppImage(c *cli.Context) error {
+
+	// check if the number of arguments are stictly 1, if not
+	// return
+	if c.NArg() < 1 {
```
piegamesde, line 47:
```suggestion
	if c.NArg() != 1 {
```

If only the first positional argument is ever used, please fail if the user provides more than one accidentally.

piegamesde:
Thank you, it works now. It'd be cool if there was a command line option to explicitly set the output file path.

srevinsaju:
> Thank you, it works now. It'd be cool if there was a command line option to explicitly set the output file path.

```bash
./mkappimage AppDir path/to/destination/FooBar.AppImage
```

> Also, I don't really understand the standalone flag. "Make standalone self-contained bundle" doesn't really explain it to me. I haven't used the flag yet the output still looks pretty standalone, so what's the deal?

I don't know either. I didn't do the coding part for the "self contained bundle". I just connected the slots to the plugs. My aim was to create a drop in replacement for the original [`appimagetool`](https://github.com/AppImage/AppImageKit) implementation, which is written in `C`. I think I should remove that flag from `mkappimage`.

src/mkappimage/mkappimage.go
srevinsaju: Improved it.

src/mkappimage/mkappimage.go
```diff
@@ -51,6 +51,15 @@ func bootstrapMkAppImage(c *cli.Context) error {
 		fileToAppImageOutput = c.Args().Get(1)
 	}
 
+    // make sure the output directory exists before continuing
+    outputDirectory := fileToAppImageOutput
+    if strings.HasSuffix(fileToAppImageOutput, ".AppImage") {
```
piegamesde, line 56: `src/mkappimage/mkappimage.go:56:8: undefined: strings`

I also don't really understand this `if` statement: Do you expect the output given to be either a directory or a file, and you assume it is a file if the path ends with `.AppImage`? That's rather weird.

My preference would be to allow the destination to only be a path to the destination file, then you could use the file API for all parts: (pseudocode) `assert(fileToAppImageOutput.exists() && fileToAppImageOutput.isFile() && fileToAppImageOutput.parent().exists())`

srevinsaju: > `src/mkappimage/mkappimage.go:56:8: undefined: strings`

Thanks. good catch.

> I also don't really understand this if statement: Do you expect the output given to be either a directory or a file, and you assume it is a file if the path ends with .AppImage? That's rather weird.

It might be. I don't have a better solution as of now.

> My preference would be to allow the destination to only be a path to the destination file, then you could use the file API for all parts: (pseudocode) assert(fileToAppImageOutput.exists() && fileToAppImageOutput.isFile() && fileToAppImageOutput.parent().exists())

Thanks, unfortunately, that doesn't help. The destination is something which would be created in future. `fileToAppImageOutput.exists()` will always give `False`, because that's something which `mkappimage` is going to create, right? So, the pseudocode will not help. 

What API would you suggest me to check if a file which is going to be written in future is a `file` or a `directory`?

srevinsaju: > src/mkappimage/mkappimage.go:56:8: undefined: strings

fixed in https://github.com/probonopd/go-appimage/pull/104/commits/f41af3fa003c98c8f0cb6e3373b01ec24d6fbbc1

piegamesde: Oh, sorry, I got things the wrong way. Let me try again: `assert(!fileToAppImageOutput.exists() && fileToAppImageOutput.parent().exists() && fileToAppImageOutput.parent().isDirectory())`

> What API would you suggest me to check if a file which is going to be written in future is a file or a directory?

You know that you won't create a directory, so you simply need to check that the path does not exist yet. And for the parent, only thing you need to know is that it exists and that it is a directory (= not a file) so that you can write in it.

If the overwrite flag is set, `!fileToAppImageOutput.exists()` becomes `fileToAppImageOutput.exists() && fileToAppImageOutput.isFile()`.

srevinsaju: > assert(!fileToAppImageOutput.exists() && fileToAppImageOutput.parent().exists() && fileToAppImageOutput.parent().isDirectory())

Agreed, that looks better. As of now, I already check if the parent directory exists before creating the appimage. Now I need to check if the parent is a directory. Thanks for your suggestion. That would do. Let me push some more commits.

srevinsaju: Yea, should do now. I have removed the `--overwrite` flag. Now the ambiguity should vanish :smile: 

Thanks for the review once again

src/mkappimage/mkappimage.go
```diff
@@ -0,0 +1,242 @@
...
+	fileToAppDir := c.Args().Get(0)
+	fileToAppImageOutput := ""
+	if c.NArg() == 2 {
+		fileToAppImageOutput = c.Args().Get(1)
+	}
+
+	// make sure the output directory exists before continuing
+	outputDirectory := fileToAppImageOutput
+	if strings.HasSuffix(fileToAppImageOutput, ".AppImage") {
+		outputDirectory = filepath.Base(fileToAppImageOutput)
```
piegamesde, line 58: I'm getting `2021/05/29 13:22:14 The specified output directory does not exist`, though it does. When the target file name does not end with `.AppImage`, this line won't be executed. Thus, the check below will try to assert the the target *file* path is a valid directory, which it of course is not.

src/mkappimage/mkappimage.go
```diff
@@ -0,0 +1,242 @@
...
+// provided as arguments exists. If yes add the current path to PATH,
+// check if all the necessary dependencies exist,
+// finally check if the provided argument, AppDir is a directly.
+// Call GenerateAppImage with the converted arguments
+// 		Args: c: cli.Context
+func bootstrapMkAppImage(c *cli.Context) error {
+
+	// check if the number of arguments are stictly 1, if not
+	// return
+	if c.NArg() < 1 || c.NArg() > 3 {
```
piegamesde, line 46:
```suggestion
	if c.NArg() < 1 || c.NArg() > 2 {
```

I'm not 100% sure, but this looks to an off-by-one mistake here.

piegamesde (commented):
I hope you don't mind my persistence on this ^^

src/mkappimage/mkappimage.go
srevinsaju: Why do you want to create an appimage which does not end with `.AppImage`?

src/mkappimage/mkappimage.go
srevinsaju: No, it looks correct to me.

srevinsaju:
> I hope you don't mind my persistence on this ^^

You are always welcome :smile:

src/mkappimage/mkappimage.go
piegamesde: File extensions aren't really worth a lot, and people might just *do* it. Btw I tested again, and it also doesn't work for me when using the correct file extension (same error message), but I can't see an obvious reason why.

This assumption may also fail in the other direction: suppose somebody accidentally creates a directory with the name of the target AppImage.

srevinsaju: > File extensions aren't really worth a lot, 

The AppImage principle does not go well with that idea, but I agree with you.

> File extensions aren't really worth a lot, and people might just do it. Btw I tested again, and it also doesn't work for me when using the correct file extension (same error message), but I can't see an obvious reason why.
>
> This assumption may also fail in the other direction: suppose somebody accidentally creates a directory with the name of the target AppImage.

I added a new commit which further refines the checks, and tries to handle the test cases as per your suggestion. 

I'm curious. Do you have a job of writing test cases? :sweat_smile: You did a really good job, and it really makes a lot of sense.

piegamesde:
It *still* doesn't work, and I'm starting to get confused. My code is equivalent to:

```sh
mkdir -p "$out"
VERSION=foobar mkappimage "AppDir" "$out/appimage-test.AppImage"
```

And I get `appimage-test.AppImage2021/05/29 18:23:40 /outfolder/appimage-test.AppImage does not exist`.

piegamesde:
Oh, I do understand: https://www.geeksforgeeks.org/filepath-base-function-in-golang-with-examples/

Could it be that you meant `filepath.Dir` instead of `filepath.Base`? I'm not into Go so I just assumed `Base` would give the parent directory, but apparently `Dir` is the correct function for this.

srevinsaju:
> Could it be that you meant filepath.Dir instead of filepath.Base? I'm not into Go so I just assumed Base would give the parent directory, but apparently Dir is the correct function for this.

Ah, yea. I am also not into Go :sweat_smile:. Let me check this one out.

src/mkappimage/mkappimage.go
piegamesde: > I'm curious. Do you have a job of writing test cases? sweat_smile You did a really good job, and it really makes a lot of sense.

Thank you :) Actually no, just university and free time stuff so far, but I'm starting to work on formal verification (of hardware) soon™ if you let that count ^^
The thing is, I've had to deal with file system foo more than once in the past and sooner or later everything that may go wrong [will blow up in your face](https://en.wikipedia.org/wiki/Murphy%27s_law). At the moment I'm only bothering you with the most basic stuff (i.e. the things that make my cursed use case work), but if you want to we can also have a little talk about symlinks and missing permissions 😈

srevinsaju: Haha, you are dope!

srevinsaju:
@piegamesde Added, https://github.com/probonopd/go-appimage/pull/104/commits/8fb36560a747c9102fc41245cb9967ab628a2d7a. Can you check out if this one works? :sweat_smile:

piegamesde:
Yes, it finally runs through :tada:. However it seems to override the output file name? It should have been `appimage-test.AppImage`, but I get `appimage-test-version-x86_64.AppImage`. The target *directory* however is fine.

srevinsaju:
> Yes, it finally runs through .


Yay, finally!! Thanks for all your help and reviews.


> However it seems to override the output file name? It should have been `appimage-test.AppImage`, but I get `appimage-test-version-x86_64.AppImage`. The target _directory_ however is fine.

Attempted a fix in https://github.com/probonopd/go-appimage/pull/104/commits/21882fbd38a4df6b17e0fa6b81725596a35ff35e

piegamesde:
It's definitely getting somewhere! All the obvious use and edge cases work for me now.

Do you know if the version is used for anything else then the file name? Because I get `Version not found, aborting. Set it with VERSION=... mkappimage` when not setting the `VERSION=`. My application isn't really versioned right now, so if I could get rid of it that'd be appreciated.

srevinsaju:
Interesting quote which I found randomly on mastodon 😄

> "Debugging is twice as hard as writing the code in the first place. 
>
> Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it."
>
> -- Brian W. Kernighan and P. J. Plauger (in The Elements of Programming Style)

srevinsaju:
> Do you know if the version is used for anything else then the file name? Because I get `Version not found, aborting. Set it with VERSION=... mkappimage` when not setting the `VERSION=`. My application isn't really versioned right now, so if I could get rid of it that'd be appreciated.

No, its not used for anything else. I can remove the requirement of `VERSION` variable, and make it optional, tomorrow.

> Another minor thing (but I'm not sure whether it is from that PR or not) is that it only searches for .png icons right now, although I am using a svg.

Oh, I will look into it tomorrow, svg's are normally acceptable for AppImages. I wonder why not. 

It looks like we are in approximately the same timezone. You are probably GMT+0200 or GMT+0100. So, good night to you too :smile:

piegamesde:
Oh the quote is totally true! That's also why I love Rust so much. You don't even need to think, the compiler will tell you what you did wrong and how to do it correctly 99% of the time so you don't even have to really think about it …

srevinsaju:
> Oh the quote is totally true! That's also why I love Rust so much. You don't even need to think, the compiler will tell you what you did wrong and how to do it correctly 99% of the time so you don't even have to really think about it …

Agreed, I love it too. I have been trying to learn it for a long time. But Rust is comparatively a harder language in my opinion, but its principle is awesome. AppImage-fork would use Rust to write the runtime (which is right now, written in C)

> Almost the same time zone (according to some browser extension guessing yours somehow). It says UTC +3 for you, and I'm UTC +2.

Haha, I guessed yours from your country, Germany 😄

probonopd:
> Also, I don't really understand the standalone flag. "Make standalone self-contained bundle" doesn't really explain it to me. 

It bundles _everything_, down to and included glibc and ld-linux. As a result, applications cannot only run on _newer_ but also on _older_ distributions than the build system.

srevinsaju:
> Do you know if the version is used for anything else then the file name? Because I get Version not found, aborting. Set it with VERSION=... mkappimage when not setting the VERSION=. My application isn't really versioned right now, so if I could get rid of it that'd be appreciated.

Removed the dependency on `VERSION` as of https://github.com/probonopd/go-appimage/pull/104/commits/3c050b73408ccc35a92621e4371dc24e5f65453f

probonopd:
Things like this are the reason why I wanted go-appimagetool to be a high-level tool. With it, I wanted to enforce AppImage best practices, so go-appimage will enforce things that the AppImage spec doesn't tightly require or specify, but which are best practices.

* I don't want to see AppImages ever again that have missing version information
* I don't want to see AppImages ever again that use svg instead of png icons
* I don't want to see AppImages ever again that are not signed
* I don't want to see AppImages ever again that have no working update information
* I don't want to see AppImages ever again that the user has renamed

Hence, go-appimagetool is made in a way that the user cannot interfere with these things, and they "just work" out of the box (if we are doing it correctly).

srevinsaju:
> Things like this are the reason why I wanted go-appimagetool to be a high-level tool. 

`mkappimage` will never be a replacement to `go-appimagetool`. It is meant to be a drop-in replacement for `appimagetool`, not `go-appimagetool`.

> I don't want to see AppImages ever again that have missing version information

The appimage spec does not specify anything with regards to embedding the AppImage's version in the desktop file.

> I don't want to see AppImages ever again that use svg instead of png icons

Personal preference is not documented. Quoting AppImage Spec Type 2 

> **MAY** contain an $APPICON.svg, $APPICON.svgz or $APPICON.png file in its root directory with $APPICON being the icon identifier as set in the Icon= key of the $APPNAME.desktop file. If present and no icon files matching the icon identifier present below usr/share/icons/hicolor, this icon SHOULD be given preference as the icon being used to represent the AppImage. If a PNG file, the icon **SHOULD** be of size 256x256, 512x512, or 1024x1024 pixels.

So, technically, `.svg`, `.svgz`, and `.png` are explicitly allowed by the AppImage Spec.

> I don't want to see AppImages ever again that are not signed

Once again, this is a personal preference. Quoting AppImage Spec:

> **MAY** embed a digital signature in the ELF section .sha256_sig. If this section exists then it MUST either be empty (filled with 0x00 padding) or contain a valid digital signature of the sha256 of the AppImage assuming the ELF section .sha256_sig being filled with 0x00 padding (why?)

This is completely optional according to the spec. According to RFC 2119

> **MAY** This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.  One vendor may choose to include the item because a
   particular marketplace requires it or because the vendor feels that
   it enhances the product while another vendor may omit the same item.
   An implementation which does not include a particular option MUST be
   prepared to interoperate with another implementation which does
   include the option, though perhaps with reduced functionality. In the
   same vein an implementation which does include a particular option
   MUST be prepared to interoperate with another implementation which
   does not include the option (except, of course, for the feature the
   option provides.)

> I don't want to see AppImages ever again that have no working update information

Quoting AppImage Spec, once again ¯\_(ツ)_/¯

> MAY embed update information in the ISO 9660 Volume Descriptor field (offset 33651). If the information in this location is not in one of the known update information formats, then it SHOULD be empty and/or be ignored


> I don't want to see AppImages ever again that have missing version information

Duplicate of the first point. 

> I don't want to see AppImages ever again that only run on the latest versions of distributions

Agreed. "latest", "old", "outdated" distributions are ambiguous terms. What according to you is an "old" distribution? Which is the oldest distro that a basic appimage should support?

> Hence, go-appimagetool is made in a way that the user cannot interfere with these things, and they "just work" out of the box (if we are doing it correctly).

Basically, it does the guess work. Guessing is not always correct, and denies the privilege to a user to explicitly configure many things. 

> If people start using the low-level/"core" tool that does not enforce the above directly (i.e., without a corresponding high-level tool that enforces these things), then we will see an influx of low-quality AppImages. 

It is probably the people's choice. AppImage already has a ton of tools, with no "recommended" tool yet. When "The AppImage Project" endorses a builder tool officially, people will eventually move to it. For example, [this project](https://github.com/probonopd/go-appimage) is not even in the [AppImage Organization](https://github.com/AppImage)

> My conclusion is that even the low-level tool should enforce these things so that no one can make AppImages that get these things wrong.

This creates a dead end for the user imo. A appimage developer should be made to understand the details, and should be responsible AppImage maintainers for their project. Any sort of guess work creates ambiguity, misunderstandings, and confusions. 

Take a look at [Snapcraft](https://snapcraft.io). They have two tools, `snap` and `snapcraft`. But most people only know `snapcraft` which can be used to build snaps. But, in real life, `snap pack` can also be used to build a `snap`. 

Well, their configuration is well designed, everything needs to be explicitly configured in the `snapcraft.yaml` build file, and any user can reproduce a snap with an identical `snapcraft.yaml` file. Their documentation is also well engineered. Perhaps its time to think about a "common" tool to recommend to the public, than to keep creating multiple tools to achieve the same endpoint. Just my 2 cents. :smile:

piegamesde:
> It bundles everything, down to and included glibc and ld-linux. As a result, applications cannot only run on newer but also on older distributions than the build system.

On which level are you talking about? I guess this is about the bootstrap binary in the AppImage ELF? Do I interpret it correctly as "static vs dynamic compilation of the AppImage launcher"? In that case, what would be a reason for *not* setting that flag?

-------

I totally understand your desire for enforcing best practises. But the thing is that there is *need* for a low-level primitive tool that does nothing more than simply bundling a folder into a self-starting ELF. This is a requirement so that others may build their own high-level abstractions on top of it.

I don't really mind opinionated tools, but I think that only specification violations should result into errors. All "not best practice" things should be lints, only resulting in warnings. (You could then add a flag comparable to `-wALL` in compilers or a `--check` to add failures to the CI or something, but hard failures should be opt-in except for spec violations).

srevinsaju:
> I don't really mind opinionated tools, but I think that only specification violations should result into errors. All "not best practice" things should be lints, only resulting in warnings. (You could then add a flag comparable to -wALL in compilers or a --check to add failures to the CI or something, but hard failures should be opt-in except for spec violations).

As of now, `mkappimage` stops making the appimage if any of the requirement as specified in the spec is not found. Any "REQUIRED" item is not "guessed" from the environment, and the user has to explicitly specify it to the tool. 

Regarding, "best practices", I agree, we should create warnings. But, you might like to check out [appimagelint](https://github.com/TheAssassin/appimagelint), an amazing tool to lint your appimage and create warnings and suggestions. It can also predict the oldest linux distro which a provided appimage can run.

probonopd:
You are right @srevinsaju the AppImageSpec is just describing the bare minimum for what is needed to make an AppImage, probably we should revise it to include the best practices:

* __SHOULD__ have an `X-AppImage-Version` field in the `.desktop` file
* __SHOULD__ have the icon as an 128x128 or 256x256 px png file
* __SHOULD__ be signed
* __SHOULD__ have working update information
* __SHOULD__ be tested to run on at least all still-supported version of the mainstream distributions such as Ubuntu, Debian, openSUSE, Fedora, etc.

srevinsaju:
> You are right @srevinsaju the AppImageSpec is just describing the bare minimum for what is needed to make an AppImage, probably we should revise it to include the best practices:

Agreed. If this makes into the `AppImageSpec`, then eventually, all the basic tools will have to support these requirements. So, a priority would be: getting the proposals to the `AppImageSpec`, so we can come to a common conclusion. ✨

If its specified as a **MUST**, then all the problems would get sorted out quickly, and there will no longer be any misunderstanding in the AppImage community, as the spec represents the ideals of "The AppImage Project", and not a single person :smile: . There will be better reason to improve for the better, together. ✨

srevinsaju:
> * **SHOULD** have an `X-AppImage-Version` field in the `.desktop` file
> * **SHOULD** have the icon as an 128x128 or 256x256 px png file
> * **SHOULD** be signed
> * **SHOULD** have working update information
> * **SHOULD** be tested to run on at least all still-supported version of the mainstream distributions such as Ubuntu, Debian, openSUSE, Fedora, etc.

Regarding the items proposed in the future spec, I would probably not be worthy to analyze them. @TheAssassin maybe :smile:

src/appimagetool/cli.go
```diff
@@ -0,0 +1,287 @@
...
+		// updateinformation: 			"" 	(empty string, we want to guess the update information from
+		//								scratch, and if we fail to guess it, then no update metadata for
+		// 								the appimage)
+		GenerateAppImage(
+			fileToAppDir, "",
+			true,
+			"gzip",
+			true,
+			"",
+			"appimagetool"
```
piegamesde, line 187: `src/appimagetool/cli.go:187:18: syntax error: unexpected newline, expecting comma or )`

piegamesde (approved):
Thank you very much for all your work on this, I don't see any major issues with this PR anymore, so it's good to go!

PR #105, opened 2020-11-24
test PR

srevinsaju:
# DO NOT MERGE

PR #106, opened 2020-11-25
ci: Add GitHub Actions CI

srevinsaju:
Possibly fixes #105 

Right now, it only works for 64-bit builds. Adding `mkappimage` at L93 would automatically configure the actions to build and release mkappimage also.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,182 @@
+name: Build
+
+on:
+  push:
+    branches: [ dev ]
+  pull_request:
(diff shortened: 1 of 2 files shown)


srevinsaju:
So far, I got `amd64/x86_64`, `arm64/aarch64` and `i686/i386` working; `arm/armv6/armv7` left to go;

wow, there is a lot of confusion in what the architectures are to be called :open_mouth:

srevinsaju:
Moved to #109

PR #107, opened 2020-11-27
DONTMERGE helpers.Args0() get's it wrong if appimaged is in PATH

eMPee584:
This a glitch I found when I hot-replaced the daemon binary in my VM experiments..
`helpers.Args0` prepends invocation path (cwd) instead of executable location
f.e. instead of /usr/local/bin/appimaged writes /home/user/appimaged into
service file

.. this is what I found to work.. sorry, just a demo patch xD

### DIFF
--- a/src/appimaged/appimaged.go
+++ b/src/appimaged/appimaged.go
@@ -88,7 +88,7 @@ var candidateDirectories = []string{
 
 func main() {
 
-	thisai.path = helpers.Args0()
+	thisai.path, _ = exec.LookPath(os.Args[0])
 
 	// As quickly as possible go there if we are invoked from the command line with a command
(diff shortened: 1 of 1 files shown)


probonopd:
Are you using our official appimaged AppImage?

eMPee584:
Just checked it, yes the official appimaged-587-x86_64.AppImage computes ai.path to `$CWD/$0`

probonopd:
Can you give an example of how to reproduce it a) when it goes wrong b) when it goes right? Thank you, this will help a lot.

eMPee584:
Sure. Put latest image at `/usr/local/bin/appimaged` (or anywhere else in `$PATH`) `cd` to f.e. `/tmp` and invoke `appimaged`

PR #118, opened 2020-12-23
Quick fix for failing builds

CalebQ42:
The build will still fail since the build script is still pulling from the main repo.

This can be reverted once [otiai10/copy](https://github.com/otiai10/copy) implements the fix.


probonopd:
> The build will still fail since the build script is still pulling from the main repo.

What is the advantage of merging this, then?

CalebQ42:
Sorry, should have been more clear. This will make the builds succeed again, but the check for the PR fails because it's pulling from the main repo.

probonopd:
In which order should I apply your PRs? And are they already ready for being merged? Thanks a lot.

CalebQ42:
I made this PR to be merged immediately. My other PR is not quite ready, but it's close.

probonopd:
Fingers crossed!

PR #129, opened 2021-02-16
check for invalid qt_prfxpath and allow user to specify it if not found

micooke:


micooke:
Sorry, I meant to merge this into my fork as part of testing. My bad

probonopd:
Hello @micooke please keep in mind that this tool is meant to run on a CI build server on which one usually cannot interactively enter something. So there should be a way to achieve the same by setting an environment variable.

micooke:
Ofcoarse! Yeah sorry about this, I was just going to test something like this on a few systems first. Ill push a PR when I fix this, with that in mind

PR #134, opened 2021-03-01
fix: improve check file exists method

srevinsaju:
https://stackoverflow.com/q/12518876

Attempts to fix an issue when the target file is a broken symlink.


probonopd:
Thanks @srevinsaju

PR #138, opened 2021-03-11
Remove trailing quotation mark from chmod

s-gbz:
Executing the original chmod commands just opens a new terminal line that is suspended indefinetely:

```
chmod +x "appimaged-*.AppImage`
>
```
It properly executes without the quotation mark though.

System: Pop!_OS 20.10


probonopd:
Thank you @s-gbz

PR #150, opened 2021-04-29
Add Version to AppImage struct

mgord9518:
Pull the version out of the AppImage desktop file, perferring X-AppImage-Version since a lot of people are lazy and don't fill out the normal Version key correctly.

### DIFF
--- a/src/goappimage/appimage.go
+++ b/src/goappimage/appimage.go
@@ -90,11 +90,7 @@ func NewAppImage(path string) (*AppImage, error) {
 	if ai.Name == "" {
 		ai.Name = ai.calculateNiceName()
 	}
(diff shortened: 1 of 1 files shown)


probonopd:
Hello @mgord9518, I'd happily consider your PR but `Version=` should not be used, not even as a fallback, because it describes the version of the format used for the desktop file itself rather than the version of the application. This comes from the [XDG Desktop Spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/) which has no standard way to describe the application version.

mgord9518:
Oh thank you! I'll resubmit without falling back to "Version", I guess I probably should've looked it up before assuming what it meant

mgord9518:
I've updated the PR

probonopd:
Thank you very much.

PR #152, opened 2021-05-25
appimaged: Allow missing notification daemon

simvux:
The previous exit branch is reached when the system is missing an notification deamon. 

Not all systems have notification daemons and there's no reason to stop execution in such a scenario.

### DIFF
--- a/src/appimaged/notification.go
+++ b/src/appimaged/notification.go
@@ -167,7 +167,7 @@ func sendDesktopNotification(title string, body string, durationms int32) {
 		log.Println("xxxxxxxxxxxxxxxxxxxx ERROR: notification:", call.Err)
 		// Sometimes we get here: "read unix @->/run/user/999/bus: EOF"
 		// means that we are not using PrivateConnection?
-        return
+		return
 	}
 
 }
--- a/src/appimaged/notification.go
+++ b/src/appimaged/notification.go
@@ -167,7 +167,7 @@ func sendDesktopNotification(title string, body string, durationms int32) {
 		log.Println("xxxxxxxxxxxxxxxxxxxx ERROR: notification:", call.Err)
 		// Sometimes we get here: "read unix @->/run/user/999/bus: EOF"
 		// means that we are not using PrivateConnection?
-		os.Exit(111)
+        return
 	}
 
 }


probonopd:
Thanks @simvux 

> Not all systems have notification daemons

Can you give an example of such a system?

simvux:
A lot of people who run tilling window managers onto bare-bone systems such as Arch or Void often skip out on notification deamons (me included). 

We're also currently trying to use appimages in an enterprise system through an [xpra session](https://xpra.org/). Of which notifications aren't included in the HTML5 client which renders windows directly rather than using a window manager.

probonopd:
> A lot of people who run tilling window managers onto bare-bone systems such as Arch or Void often skip out on notification deamons (me included).

Those are non-supportable because no 2 systems are exactly the same. Issues with such self-made ad-hoc mixtures of components cannot be reproduced easily. Which will lead to bug reports. So even if I merged this PR you'd probably run into other issues.

So for now, a notification system is a dependency of go-appimaged. How else should it communicate with the user?

simvux:
@probonopd 
AppImage's tagline is "Linux Apps that run anywhere" yet you don't merge a pull request which changes one line of code to make this appimage tool supported by 10% more systems, while also providing no reason for why this change is in any way a bad thing. 

A daemon which does work in the background requiring a notification daemon to not crash is just bad code. 

I'd like to ask you to strongly reconsider.

probonopd:
> A daemon which does work in the background requiring a notification daemon to not crash is just bad code.

Agree. We should make it print out an easy-to-understand message saying that a notification daemon was not found on the system, and exit cleanly. A PR that does this would be appreciated since the Live ISOs I am developing this on all come with a notification daemon as part of the system.

simvux:
> We should make it print out an easy-to-understand message saying that a notification daemon was not found on the system, and exit cleanly. 

I see this as an direct downgrade from just printing an error message and *continuing execution*. Exiting from an error like this makes *absolutely no sense*.

probonopd:
Why? How else can we enforce correct operation of go-appimaged, including it to show messages to the user? (go-appimaged without a possibility to show messages to the user is half-broken)

probonopd:
To be clear, I am not against merging this. I am just pointing out the larger issue at hand. Utter fragmentation of the "Linux" platform.

> log.Println("xxxxxxxxxxxxxxxxxxxx ERROR: notification:", call.Err)

Let's put in a nice error message at least, shall we?

simvux:
Thanks for the merge. 

> Why? How else can we enforce correct operation of go-appimaged, including it to show messages to the user? (go-appimaged without a possibility to show messages to the user is half-broken)

The old code made it fully-broken rather than half-broken. And I wouldn't even consider it half-broken since notifications is by no means 50% of the application's purpose. 

> Why not just install a notification daemon on your system, like most mainstream distributions do out of the box?

This is no reason to make your application a landmine of possible crashes.

PR #156, opened 2021-06-28
appimaged: use v2.x of AppImageUpdater if available.

antony-jr:
The new releases of AppImageUpdater uses ```latest``` instead of ```continuous```. So search for new versions of AppImageUpdater too. **Also starting from v2.x the updater uses xdg-su instead of sudo.**

**TL;DR:** Use v2.x of AppImageUpdater if found.

I've tested this on my local machine and it seems to work fine.


probonopd:
Thank you @antony-jr

Issue #157, opened 2021-07-02
Does not recognize lowercase extensions


omvsj:
I've just switched from old version of appimaged to go-appimage and my applications weren't recognized.
I should note it was hard for me to tell what was happening, since it is a new version and do not work like previous version:

1) go-appimage doesn't have --install or --uninstall switches
2) go-appimage doesn't move itself to other location on installation
3) there's no indication on how to install it
4) there's no indication of what folders are being watched

I rann appimaged-646-x86_64.AppImage and I couldn't know if it was properly installed and if it was "watching" Applications folder. I do noticed go-appimage didn't move to a new location and kept in the same folder it was started. I moved it to Applications folder and a notification appeared showing something like "1 application added". In that moment I knew it did were installed, it did was running and it did was watching Applications folder but it didn't registered my applications. I move my applications outside Applications folder and inside again but nothing happened. I noticed capitalization of extensions were different. I changed them to match "AppImage" and they were recognized.

Although all of this are "minor" problems, newbies could have a hard time trying to get it work. So, I suggest putting this into the Readme:
1) Download appimaged-646-x86_64.AppImage to Applications folder. To install just run it and do not remove it. Users may think that they can remove it if it is already installed.
2) Tell users what folders are being watched
3) Tell users that all AppImages extensions must be properly capitalized

probonopd:
Thanks for your correct observations @omvsj. Would you mind sending a pull request for the README?

omvsj:
I had never put a "pull request" but sure, I'll do it.

PR #158, opened 2021-07-06
Update README.md

omvsj:
As commented on https://github.com/probonopd/go-appimage/issues/157#issue-935975668

This is a proposal to let new users better know how to install, uninstall and name AppImage files correctly to get service working.

I only doubt if it is convenient to delete the downloaded appimaged file or leave it where it was downloaded when user is uninstalling it.


probonopd:
Thanks @omvsj

PR #159, opened 2021-08-01
Attempt to convert SVG icons to PNG

mgord9518:
Should take weight off of integration time, the library attempts to use existing software on the dev's machine (eg: Inkscape, ImageMagick, rsvg-convert) to convert SVG icons to PNG format before saving as .DirIcon


probonopd:
Thanks @mgord9518

PR #160, opened 2021-08-05
Makes AppImage search case insensitive

bflanagin:
This should make the search for appimage files case insensitive.


probonopd:
Thanks @bflanagin

PR #166, opened 2021-09-04
Build script fix & code cleanup

CalebQ42:
The most important part of this is changes to the build script to make it work properly again. There were several issues pertaining particularly to building `mkappimage` that seemed to have never been tested.

- Updated project to go 1.17. Go 1.13 hasn't been officially supported for a bit now, plus the past few releases have made several improvements to performance.
- Remove generated bindata.go in `appimaged` and replace it with the embed package. This is partially just a personal preference of not seeing a wall of symbols in any part of my code, not to mention if we need to make any changes, such as adding more files, it will be a lot easier.
- Made some tweaks to `mkappimage`, such as removing a bunch of unnecessary code and an unnecessary cgo import which was causing some issues. I'll probably look at `mkappimage` a bit more at some point since it's based on an old version of `appimagetool` that was not making use of `goappimage` and instead using `unsquashfs`.
- Some code cleanup that was suggested by [gopls](https://github.com/golang/tools/blob/master/gopls/README.md). This mostly consisted of removing a bunch of `== true` statements and some unused variables.


CalebQ42:
These were mostly just things I noticed/needed to do before looking deeper to figure out some of the issues that some people have been having.

probonopd:
Thank you very much @CalebQ42. Do you think you could add a GitHub Action to build and upload each commit?
That would be tremendous, since Travis CI is broken and will not come back in its old .org form.

CalebQ42:
> Thank you very much @CalebQ42. Do you think you could add a GitHub Action to build and upload each commit?
> That would be tremendous, since Travis CI is broken and will not come back in its old .org form.
> 
> I am a bit hesitant to merge anything before we have a working CI again.

I can definitely work on this. If it's alright, I'll probably also make some changes so the build script works well locally. I personally have to make a few changes to it each time I want to test something.

probonopd:
> I can definitely work on this.

All the power to you! Thank you very much for looking into it :+1:

CalebQ42:
So it looks like GitHub Actions is going to end support for Ubuntu 16.04 very soon, but since it's EOL as of April, we probably should move to 18.04 as the standard build environment anyway.

probonopd:
Running into an issue:
https://github.com/probonopd/go-appimage/runs/3514798957?check_suite_focus=true

CalebQ42:
Sorry about that, I was working to debug it, but the actual build wasn't even starting for me and I had to go to sleep (it actually never started and I had to manually stop it after I woke up). I'll submit a new PR when I get all the kinks worked out.

probonopd:
No problem, take your time. Thanks!

PR #170, opened 2021-09-10
GitHub actions

CalebQ42:
I ended up making some major changes to the build script to make it a bit more useful. You can now specify which architectures and programs you want to build, which should be nice for testing. Github Actions should also be fully working now, though the builds for arm and arm64 takes a significant amount of time ([the latest compile](https://github.com/CalebQ42/go-appimage/actions/runs/1221707619) took 23 minutes) since Github Actions doesn't have native support for arm.

Commits are a bit messy due to not using bash in forever and never using Github Actions or Docker so I had to done tons of iterating.


probonopd:
Thanks @CalebQ42

probonopd:
Wow, that looks good :+1: 

image

Regarding aarch64 being _that_ slow, do you think we could make use of the Go cross-compile feature and put together the AppImage on a normal x86_64 system?

CalebQ42:
There's definitely some merrit there, though from what I understand we still need a native system to build the actual AppImage, since we need to use that architecture's AppImageTool. Admittedly, there's most likely a way to use docker in the build script to do this, but I've never used docker before and the docker parts of this system were basically just copied and pasted from a forum thread.

In theory, future builds SHOULD be faster since some of it is cached for the next build.

I probably will continue to look into this as I do more work on the actual codebase as waiting for the arm build over so many small iterations as I ironed out the last few bugs was a bit painful.

probonopd:
>  from what I understand we still need a native system to build the actual AppImage, since we need to use that architecture's AppImageTool

It should be relatively easy to change appimagetool so that it can use runtimes for other architectures:

https://github.com/probonopd/go-appimage/blob/8ecb9435138cbdad1038da837d7a8a9459834fff/src/appimagetool/appimagetool.go#L368-L372

Maybe worthwhile?

CalebQ42:
This looks like it'll work. It shouldn't take too long to get it implemented too.

probonopd:
Hello @CalebQ42, we need to upload the binaries to the `continous` release in GitHub Releases.

Similar to here:

https://github.com/probonopd/custom-tasmota/blob/239db852ff886db9d2500c7f039a50c75886b3e6/.github/workflows/ci.yml#L47-L57

Can you add this please?

PR #175, opened 2021-09-14
Build Upload

CalebQ42:
Added uploading of builds when pushing to the master branch


probonopd:
Wonderful!

probonopd:
Do you think we can add 646 to the number we export as `$VERSION`? This way, updates from the Travis CI generated AppImages to the GitHub generated ones would be seamless, and the "versioning" would be consistent.

Issue #151, opened 2021-05-04
Continuous build should not contain version number


sjdrc:
I have scripts that fetch appimagetool, but in the continuous build on this repo, you put the version number into the filename, making it impossible to reliably download the image.

Can you please either omit the version number from the filename, or do actual versioned releases with stable URLs.

sjdrc:
@probonopd I see that the latest builds now have `-1-` in the filename. Is this name going to stay the same going forwards?

probonopd:
Works as intended again.

PR #176, opened 2021-09-15
Fix build number

CalebQ42:


sjdrc:
Would it be possible to just omit the version number from the filename altogether? With a continuous release this makes it incredibly annoying to have our CI pipelines fail randomly when there is a new version released. There is no reliable URL to download build artifacts from. Please see #151

probonopd:
@sjdrc

This is what we use:

```
wget -c https://github.com/$(wget -q https://github.com/probonopd/go-appimage/releases -O - | grep "appimagetool-.*-x86_64.AppImage" | head -n 1 | cut -d '"' -f 2)
chmod +x appimagetool-*.AppImage
```

PR #181, opened 2021-12-03
Add permissions and update information to AppImage struct

mgord9518:
Give AppImages a way to request simple permissions.

### DIFF
--- a/internal/helpers/desktopfile.go
+++ b/internal/helpers/desktopfile.go
@@ -34,3 +34,12 @@ func CheckDesktopFile(desktopfile string) error {
 
 	return nil
 }
+
+// Converts a multi-item INI value into a slice
+// eg: `foo;bar;` becomes []string{ "foo", "bar" }
+func SplitKey(str string) []string {
+    str = strings.ReplaceAll(str, "；", ";")
+    f := func(c rune) bool { return c == ';' }
+
+    return strings.FieldsFunc(str, f)
+}
--- a/internal/helpers/updateinformation.go
+++ b/internal/helpers/updateinformation.go
@@ -1,6 +1,8 @@
 package helpers
 
 import (
+	"bytes"
+	"debug/elf"
 	"errors"
 	"net/url"
 	"strings"
@@ -121,4 +123,31 @@ func ValidateUpdateInformation(updateinformation string) error {
 
 func getChangelogHeadlineForUpdateInformation(updateinformation string) string {
 	return ""
+}
+
+// Taken and modified from
+// <https://github.com/AppImageCrafters/appimage-update/blob/945dfa16017496be7a3f21c827a7ffb11124e548/util/util.go>
+func ReadUpdateInfo(appImagePath string) (string, error) {
+	elfFile, err := elf.Open(appImagePath)
+	if err != nil {
+		return "", errors.New("file not found")
+	}
+
+	updInfoSect := elfFile.Section(".upd_info")
+	if updInfoSect == nil {
+		return "", errors.New("ELF missing .upd_info section")
+	}
+	
+	sectionData, err := updInfoSect.Data()
+	if err != nil {
+		return "", errors.New("unable to read update information from section")
+	}
+
+	str_end := bytes.Index(sectionData, []byte("\000"))
+	if str_end == -1 || str_end == 0 {
+		return "", errors.New("no update information found")
+	}
+	updInfoStr := string(sectionData[:str_end])
+
+	return updInfoStr, nil
 }
--- a/src/goappimage/appimage.go
+++ b/src/goappimage/appimage.go
@@ -31,10 +31,21 @@ type AppImage struct {
 	Desktop *ini.File
 	Path    string
 	// updateInformation string TODO: add update stuff
-	Name      string
-	Version   string
-	offset    int64
-	imageType int
+	Name         string
+	Version      string
+	Permissions *Permissions
+	offset       int64
+	imageType    int
+}
+
+// Simple, Android-like permissions an AppImage can request
+// Note that these are ONLY for external sandbox implementations to use, no
+// sandboxing is done by default
+type Permissions struct {
+	Level     int    // How much access to system files
+	Files   []string // Grant permission to access files
+	Devices []string // Access device files (eg: dri)
+	Sockets []string // Use sockets (eg: x11, network)
 }
 
 // NewAppImage creates an AppImage object from the location defined by path.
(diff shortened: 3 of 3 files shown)


CalebQ42:
These changes seem to be only for your [aisap](https://github.com/mgord9518/aisap) sandboxing library instead of something that should be added into the main `goappimage` library. Since the official way to sandbox AppImages is via Firejail, I don't see any particular reason why this should be a part of this library and not a part of your `aisap` library unless @probonopd would want it to be a more integrated sandboxing solution. What would probably work better would changing `aisap` to extend the `goappimage.AppImage` type like:
```
type AppImage struct{
    goappimage.AppImage
    Permissions *Permissions
}
```
or if you need to overwrite some of the functions, something like this might be more appropriate:
```
type AppImage struct{
    ai goappimage.AppImage
    Permissions *Permissions
}
```

mgord9518:
My intention is to create a basic standard so that other sandboxing implementations can have an idea of the permissions an AppImage requires without always having to make custom profiles. I started aisap because I encountered a lot of issues using (primarily lesser known) AppImages through firejail, I believe this solution could also benefit it (and possibly others in the future) if they chose to use them.

@probonopd [mentioned](https://discourse.appimage.org/t/feature-request-allow-telling-appimage-mount-to-use-a-specific-directory/2445/4) that aisap was roughly what he imagined for sandboxing in go-appimage, so I figured to try to add the most basic element and keep it as generic as possible. If need be, I can resubmit the PR only keeping the UpdateInfo changes

probonopd:
Do we really need to introduce AppImage-specific desktop file keys for this, isn't there some standard? How do Snap and Flatpak communicate the permissions needed? I would like to avoid this being an `X-AppImage` thing. AppImage, after all, is just a self-mounting disk image that executes whatever the author has put inside.

mgord9518:
I looked all over for some sort of existing standard and wasn't able to find anything, so if it does exist it's pretty well hidden. Flatpak and Snap seem to store permissions fundamentally differently, [Flatpak's](https://github.com/flatpak/xdg-desktop-portal/wiki/The-Permission-Store) being binary and non-human readable for some reason. Going directly to XDG to ask is a good idea

mgord9518:
@probonopd I was able to get in contact with the mailing list and they say there's no existing spec for this. Here's the archive of my original [post](https://lists.freedesktop.org/archives/xdg/2021-December/thread.html), I'm not sure when the reply will be available in the archives but it should be fairly soon being that it only took a day for the first post to pop up.

probonopd:
I think the XDG people are pushing the idea of using "portals" 

> Portals originated from the Flatpak project

and

> portals require applications to use a new API in order to access resources

which I think is less than ideal as it would require additional work for all existing applications, and doesn't allow existing binaries to be sandboxed in this way.

Instead, I think applications should just specify what permissions they request, and the sandbox would have to enforce that applications only get what they requested. Users could view the requested permissions upfront and decide whether they want to use the app based on that information.

mgord9518:
@probonopd Yeah, it definitely seems that way. If not forcing, they simply don't appear to care about non-Flatpak solutions.

How do you think AppImages specify their required permissions? I think extending the desktop file is a pretty good way to go about it, but maybe it could just be named more generically as to make it an option for other formats like native distribution as well? Instead of `X-AppImage-Required-Permissions` it could just be `X-Required-Permissions`. This is extremely easy to interpret and implementing it in any sandbox should also be pretty easy.

probonopd:
It should **not** be something AppImage specific. Something like `X-Required-Permissions=...` sounds good to me. In the ideal world, a proposal would be made and would be discussed on the XDG mailing list, so that it would become an official standard and the `X-` could be dropped. Maybe the Flatpak-centric stuff tends to end up in XDG simply because no one else makes proposals...

mgord9518:
@probonopd I tried to contact the mailing list, I only got one reply and they just said that no such standard exists. For some reason their reply still hasn't been added to the archive, but it was "I have read the spec and i do not see anything to sandbox stuff autho a good sandboxer would be firejail". 

I don't think they any plans to add something like this, maybe if `X-Required-Permissions` spreads outside the AppImage world, it *could* eventually get adopted, similar to `XDG_STATE_HOME`, which I believe was actively used by a good amount of programs years before being officially standardized.

The idea I have is 3 basic levels. 1 is the most lenient and gives access to all device files, all of /sys, all of /usr and /etc and only really restricts access of the user's files. This would be used for stubborn applications that refuse to be sandboxed to higher levels for some reason. Level 2 would restrict devices, only giving access to those requested and their related /sys files, /usr is also restricted to a degree, but themes, ca-certs, etc are allowed. It's the go-to for GUI applications. Level 3 restricts practically everything outside of the specific requested files, and is intended for use with CLI applications, but completely self-contained GUI applications would also be able to use it.

Besides the level: devices, files and sockets are requested as a list, and they all get applied on top of the base level. This is how I have it implemented in aisap currently

probonopd:
@mgord9518 are you still interested in getting this kind of functionality into appimaged? I'd be.

mgord9518:
@probonopd Absolutely! Maybe an environment variable could enable it?

probonopd:
Or the presence of a compatible sandbox on the device.

mgord9518:
@probonopd that's an option, but I still think it should be opt-in. Otherwise, anyone who has Flatpak on their system (or bwrap on its own) could suddenly experience incompatibilities with certain AppImages or just new, unexpected behavior (eg: only being able to access certain directories in their web browser) upon updating appimaged.

There's also 2 ways off the top of my head that aisap could be added to appimaged. Option one is just to drop aisap in, replacing go-appimage in appimaged as aisap re-implements its API. This would be the easiest option and is how I originally designed aisap to be used as it extends go-appimage and it's hard to use them together in the same project. Option 2 would be merging aisap's sandboxing capabilities into go-appimage. I believe this would be the better option overall, but it expands the scope of go-appimage.

I'd be interested to hear which one you prefer, or of course any possible other ways to do it.

probonopd:
> I still think it should be opt-in. Otherwise, anyone who has Flatpak on their system (or bwrap on its own) could suddenly experience incompatibilities with certain AppImages or just new, unexpected behavior 

Agree.

By the way, is there any chance we could use Firejail instead of bwrap? Reason: The Firejail developers actually care about AppImage. And I don't want to use dependencies where upstream the AppImage use case.

mgord9518:
@probonopd it's certainly possible, I'd probably just have to learn all of Firejail's counterparts to the bwrap permissions aisap is currently using. I wouldn't remove support for bwrap, but I can't think of any reasons not to add Firejail support

PR #189, opened 2022-02-05
Add `UpdateInfo` and small clean up

mgord9518:


probonopd:
Thanks @mgord9518. Is there a reason you didn't want to pull in https://github.com/AppImageCrafters/appimage-update as a dependency and rely on that implementation?

@azubieta what would be your recommendation?

mgord9518:
@probonopd Honestly there's no reason, I guess it would be a good way to kill 2 birds with 1 stone by also integrating AppImageUpdate into go-appimage. I can modify the PR if you feel it would be the better option

probonopd:
Let's see what @azubieta thinks and whether https://github.com/AppImageCrafters/appimage-update will still be maintained.

azubieta:
The zsync2 implementation on appimage-update is incomplete. Right now it's only capable or reusing a very limited part of the file delta and in some cases it performs worse that a regular download. It needs to be polished.

probonopd:
Thanks @mgord9518. Maybe we will get to an integrated updater in go-appimage at some point in time.

Issue #183, opened 2021-12-05
Allow generated desktop entries to be customised


jamerst:
I have an appimage which requires some additional commands and arguments in order to launch successfully on my system. However, every time the daemon restarts my modifications to the desktop entry are overwritten.

It would be very useful if appimaged could either preserve modifications somehow, or if there was a way to customise the desktop entry for a specific appimage.

probonopd:
Can you give a concrete example please?

Please also check what is inside the AppImage in its desktop file as the `Exec=` line.

jamerst:
The file is generated with:
```
Exec=/home/james/.local/bin/appimaged-650-x86_64.AppImage wrap "/home/james/.local/bin/Plexamp-3.8.2.AppImage"
```

I have to change it to: 
```
Exec=env LD_PRELOAD=/usr/lib64/libcrypto.so.1.0.2o /home/james/.local/bin/appimaged-650-x86_64.AppImage wrap "/home/james/.local/bin/Plexamp-3.8.2.AppImage" "--no-sandbox"
```

The desktop file inside the appimage already has the `--no-sandbox` argument, but I still have to add it to the generated file. Unfortunately I don't have control over the contents of the appimage, the bug this is a workaround for has existed for well over a year so I don't envision it being fixed any time soon.

If customisations aren't easily possible, I guess some kind of blacklist system would suffice so that I can manage the desktop entry for this specific appimage manually.

probonopd:
Would https://docs.appimage.org/user-guide/troubleshooting/electron-sandboxing.html#configure-unprivileged-sandboxes help maybe?

jamerst:
The `unprivileged_userns_clone` kernel parameter is a Debian patch, so it's not available for non-Debian distros. I use Fedora, so I can't use it unfortunately.

The `--no-sandbox` argument seems to be a common workaround, it's used in most of the other Electron-based appimages I have installed. It works fine for those other applications though, it's only Plexamp that needs it to be added manually.

The bug that requires LD_PRELOAD is specific to Plexamp I believe, I haven't seen the issue anywhere else.

probonopd:
> The unprivileged_userns_clone kernel parameter is a Debian patch, so it's not available for non-Debian distros. 

Are you really sure about this? I thought all distributions use the same _kernel_ - Linux.

jamerst:
Whilst all distros use the Linux kernel, many distros will apply their own customisations on top of the kernel in the form of patches.

[...]

PR #193, opened 2022-04-01
Fix #192

### LINKED ISSUE #192: appimaged desktop files: Arguments are not passed correctly
Steps to reproduce:

1) Download an appimage
2) let appimaged generate a desktop file
3) cause arguments to be passed to the application: e.g. download https://freetubeapp.io/ which registers `MimeType=x-scheme-handler/freetube` to handle `freetube://` urls.
4) Open a freetube URL, e.g. `freetube://https://www.youtube.com/watch?v=dQw4w9WgXcQ` with `xdg-open`

Expected behaviour: FreeTube opens and starts playing rickroll.

Actual behavior: FreeTube opens and doesn't play anything.

The generated desktop file contains this exec line:

    Exec=/home/user/Applications/appimaged-650-x86_64.AppImage wrap "/home/user/Applications/freetube.AppImage"

but it should really be

    Exec=/home/user/Applications/appimaged-650-x86_64.AppImage wrap "/home/user/Applications/freetube.AppImage" %U

so that the arguments are passed correctly.

CalebQ42:
Fixes #192 by adding arguments that are present in the AppImage's desktop file to the generated desktop file. Also a partial fix for #183.

### DIFF
--- a/src/appimaged/desktop.go
+++ b/src/appimaged/desktop.go
@@ -93,7 +93,12 @@ func writeDesktopFile(ai AppImage) {
 	// so that renaming the file in the file manager results in a changed name in the menu
 	// FIXME: If the thumbnail is not generated here but by another external thumbnailer, it may not be fast enough
 	time.Sleep(1 * time.Second)
-	cfg.Section("Desktop Entry").Key("Exec").SetValue(arg0abs + " wrap \"" + ai.Path + "\"") // Resolve to a full path
+	add := ""
+	args, err := ai.Args()
+	if err == nil {
+		add = " " + strings.Join(args, " ")
+	}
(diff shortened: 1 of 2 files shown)


CalebQ42:
Actually, should only field codes (such as `%U`) be included since other arguments should already be passed to the AppImage's AppRun by the AppImage itself?

probonopd:
Thanks @CalebQ42.

> But then in https://github.com/probonopd/go-appimage/issues/183 it's mentioned that --no-sandbox is needed in both places, which seems a bit odd.

I think I can explain this, if memory serves me right:

So far, we have been disregarding arguments in desktop files in `AppRun.c`. Instead, so far we have only used the first argument from the `Exec= `line and have appended any arguments given on the comand line to the AppImage. I think we did so because we were too lazy to implement all the complicated %`U`, `%u`, `%F`, `%f `stuff in `AppRun.c`. Let's also keep in mind that not all AppImages use `AppRun.c` at all.

PR #194, opened 2022-04-11
Build with 1MiB block size

mgord9518:
Using a bigger block size than mksquashfs's default improves read speed and can produce smaller AppImages as well


probonopd:
Thanks @mgord9518

PR #196, opened 2022-04-17
Add --runtime-file CLI option to mkappimage

mgord9518:
Feature parity with https://github.com/AppImage/AppImageKit/blob/master/src/appimagetool.c

Allows embedding an external runtime in the AppImage instead of using those contained in mkappimage


probonopd:
Thanks @mgord9518. Very useful! Appreciate it. Happy Easter.

mgord9518:
Thanks, you too!

PR #203, opened 2022-05-07
Allow to statically link Go programs, even with cgo

probonopd:
Statically compile Go programs, even with cgo, using musl libc, like this:

```
wget -c -q http://www.musl-libc.org/releases/musl-1.1.10.tar.gz
tar -xvf musl-*.tar.gz
cd musl-*/
./configure
make -j4
sudo make install
CC=/usr/local/musl/bin/musl-gcc go build --ldflags '-linkmode external -extldflags "-static"' hello.go
```

Credits: https://honnef.co/posts/2015/06/statically_compiled_go_programs__always__even_with_cgo__using_musl/

We can probably get around compiling musl libc by using an Alpine Linux container rather than an Ubuntu one in the first place.

### DIFF
--- a/.github/workflows/build_pr.yaml
+++ b/.github/workflows/build_pr.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-all:
     name: Build for all architectures
-    runs-on: ubuntu-18.04
+    runs-on: alpine-latest
     steps:
     - uses: actions/checkout@v2
     - name: Set up Go
--- a/.github/workflows/build_pr.yaml
+++ b/.github/workflows/build_pr.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-all:
     name: Build for all architectures
-    runs-on: alpine-latest
+    runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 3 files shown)


probonopd:
**Tradeoff:** Using musl libc may result in us being unable to use mDNS/Zeroconf.
So possibly we may not want to use it for `appimaged`.

> mDNS and alternate hostname database backends
The inability to use mDNS (a multicast-DNS-based zero config system) with musl has been raised as an issue by users in the past. On glibc, using mDNS is accomplished with NSS; obviously musl does not have (or want) NSS.
>
> In principle, however, musl is fully extensible to use alternate hostname database backends in place of normal DNS. All that’s needed is a daemon that runs on localhost, speaks DNS, and translates the requests to whatever backend is needed. However it’s unclear whether there are any existing tools of this form. Developing one, adapting an existing DNS proxy program, or documenting how to setup an existing program that’s already capable could be a nice future project.

probonopd:
Continued in https://github.com/probonopd/go-appimage/pull/207

PR #211, opened 2022-06-15
Added a check for broken symlinks, should fix probonopd#210

dandreossi:
If in `xdg.xdg.DataHome + "/applications/"` is present an invalid symlink, the appimaged wrapper crash, often without starting the application.
Issue should be solved by adding a check if the symlink is valid and, if it's valid, the target file will be parsed.


probonopd:
Thanks @dandreossi

PR #219, opened 2022-09-08
Build all runtimes into AppImage

mgord9518:
Feature parity with appimagetool, allows "cross-building" AppImages of a different arch than the host


probonopd:
If all archs are put into the AppImage, can they actually be used from there?

mgord9518:
Yeah it works

probonopd:
Thanks @mgord9518 - that was an easy but valuable fix thren!

PR #225, opened 2022-10-08
Small fix

CalebQ42:
* Fixed deleted AppImages not being removed.
* Updated go.mod to use go 1.19, and updated dependencies.
* Removed instances if `io/ioutil` since it's depreciated.
* `go fmt` also fixed some formatting when doing above.


CalebQ42:
Fixed an issue in the Yuzu appimage where the desktop file's symlink started with ./ which, according to `io/fs` is an invalid path with squashfs v0.6.2.

probonopd:
Thanks @CalebQ42

PR #227, opened 2022-10-14
Use predictable file names for the continuous release

quyykk:
On each update to the continuous build the version number gets incremented. This is not ideal because it makes it impossible to predictably always download the latest continuous build.

This PR changes the version in the filename to continuous for the continuous release, so that the filenames are always the same.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,6 +17,8 @@ jobs:
   build-all:
     name: Build for all architectures
     runs-on: ubuntu-18.04
+    env:
+      VERSION: continuous
     steps:
     - uses: actions/checkout@v2
     - name: Set up Go
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -206,7 +206,6 @@ done
 # Export version and build number
 if [ ! -z "$GITHUB_RUN_NUMBER" ] ; then
   export COMMIT="${GITHUB_RUN_NUMBER}"
-  export VERSION=$((GITHUB_RUN_NUMBER+646))
 else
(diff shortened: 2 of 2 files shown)


probonopd:
Each AppImage shall have the version in its name so that people can keep around multiple versions easily.

To download the latest continuous build in scripts, you can use something like

```
wget -c https://github.com/$(wget -q https://github.com/probonopd/go-appimage/releases -O - | grep "appimaged-.*-x86_64.AppImage" | head -n 1 | cut -d '"' -f 2)
```

Ideally we would have a static URL that would redirect to always the latest download URL, but I haven't figured out a way to do this on GitHub Releases.

quyykk:
> Ideally we would have a static URL that would redirect to always the latest download URL

There's https://github.com/probonopd/go-appimage/releases/tag/continuous which links to the continuous release directly, but I don't think you meant that.

Unfortunately your snippet doesn't work because GitHub loads the assets with JS, and so they're not available when using wget. I don't think there is a way to get the latest download URL directly without predictable filenames, except of course by doing GitHub API calls but that's not trivial.

As an alternative, could you periodically create new tags (i.e. releases) so that there is a stable download URL?

probonopd:
Indeed they seem to have changed something, so https://github.com/probonopd/go-appimage/releases/expanded_assets/continuous needs to be used now.

probonopd:
Tested, working:

```
wget -c https://github.com/$(wget -q https://github.com/probonopd/go-appimage/releases/expanded_assets/continuous -O - | grep "appimaged-.*-x86_64.AppImage" | head -n 1 | cut -d '"' -f 2)
```

Issue #48, opened 2020-08-01
Cannot see volumes come and go on Kubuntu 20.04


probonopd:
![image](https://user-images.githubusercontent.com/2480569/89103281-2986a580-d400-11ea-9679-c2314117a972.png)

probonopd:
org.kde.Solid.Device messages are there, but apparently the way we determine whether we can see volumes come and go does not work on Kubuntu 20.04.

```
me@host:~$ dbus-monitor 2>&1 | grep -i setupDone
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb2',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb2',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb1',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb1',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb2',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb2',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb1',interface='org.kde.Solid.Device',member='setupDone'"
   string "type='signal',path='/org/kde/solid/Device__2Forg_2Ffreedesktop_2FUDisks2_2Fblock_5Fdevices_2Fsdb1',interface='org.kde.Solid.Device',member='setupDone'"
[...]

PR #233, opened 2022-11-29
Fixed can't see volumes on kde (#48)

CalebQ42:
Removed syslog from systemd service file since it's depreciated.


CalebQ42:
Tested on Kubuntu 22.10 in a VM

probonopd:
Thank you very much @CalebQ42

PR #234, opened 2022-12-01
Wide refactoring

psydvl:
- Separate `mkappimage` and `appimagetool` files
- Move some functions in separate packages or helpers
- Move `GenerateAppImage` function into `goappimage` package
- Some fixes
- Regenerate `ExcludeLibraries`


probonopd:
Hi @psydvl,
thanks for working on go-appimage.
I understand "Follow [Standard Go Project Layout](https://github.com/golang-standards/project-layout)", but what else do you want to achieve? In other words, _why_ do you think this should be refactored?

psydvl:
Mostly written in commit messages, but general points: 
- Symlinking appimagetool files in mkappimage is bad decision because of logic complexity 
- os.Exit bad practice due two things: 
  1. defer calls declined by it; can be solved by panic instead (will call all defer and raise exit code 2)
 . 2. "_go way_" is to raise errors from function to the highest level, keeping handle it as long as possible 
- common function should be raised in separate simple packages when possible 
- there are some recommendation to keep packages in cmd (other words cli utilities) slim as possible moving most function in other folders (that's why it is still draft) 

Of course, I may be wrong about some changes and open to discuss any 

Also, some fixes done due 
- goimports auto formatting 
- gofmt/gofumpt auto formatting 
- default linters, from golangci-lint, recommendation 

Steps I want to do in next commits: 
- [ ] move cmd code to separate packages where its possible 
- [ ] improve error handling (maybe with [multierr](https://github.com/uber-go/multierr) or [go-multierror](https://github.com/hashicorp/go-multierror))
- [ ] leave no warns on linters check 


Inspired by: 
https://go.dev/doc/effective_go 
https://google.github.io/styleguide/go

probonopd:
Yes, let's not introduce the need for a makefile.

Issue #112, opened 2020-12-05
FUSE filesystem instead of copying around desktop files


probonopd:
Right now we are copying desktop files (and other similar files, such as icons) into the running system when applications are integrated.

This has the downside that in case appimaged is killed/removed from the system, those extraneous files are not removed from the system.

If we would use a FUSE-mounted filesystem for those, then if the process that mounts the FUSE filesystem got killed, those files would magically disappear. Also we would not have to copy anything anywhere, which I think would be much cleaner and nicer.

probonopd:
https://godoc.org/github.com/hanwen/go-fuse/fs

probonopd:
Maybe even better?
https://github.com/bazil/fuse

> It is a from-scratch implementation of the kernel-userspace communication protocol, and does not use the C library from the project called FUSE.

probonopd:
https://github.com/jacobsa/fuse

grahamperrin:
I sort of don't care how it's done, but generally, with a passion: 

> … instead of copying around …

Yes. Oh, yes.

Issue #230, opened 2022-11-04
Go get deprecated, build instructions out of date


gamedevsam:
I attempted to follow instructions here: https://github.com/probonopd/go-appimage/blob/master/src/appimaged/README.md#building

And ran into the following error:
```
go get github.com/probonopd/go-appimage/src/appimaged
go: go.mod file not found in current directory or any parent directory.
	'go get' is no longer supported outside a module.
	To build and install a command, use 'go install' with a version,
	like 'go install example.com/cmd@latest'
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
```

gamedevsam:
For anyone interested in building this yourself, it's as simple as:

```
git clone https://github.com/probonopd/go-appimage
cd go-appimage
./scripts/build.sh -a amd64
```

CalebQ42:
It looks like the individual tool's READMEs have just been neglected for a bit. It would probably be best if you submitted a PR to fix them.

As a note, you only have to specify `-a` in the build script if you're building for non-native architectures (and the build script currently only works on amd64). Also the individual READMEs should only build for the current tool so it would be more correct as `scripts/build.sh appimaged`.

parkerlreed:
zig download gives a 403 forbidden, halting the building process.

```
(1)(deck@steamdeck go-appimage)$ ./scripts/build.sh -a amd64
+ '[' 2 -gt 0 ']'
+ case $1 in
+ BUILDARCH=(${2//,/ })
+ shift
+ shift
+ '[' 0 -gt 0 ']'
+ '[' '!' -z '' ']'
++ date +%Y-%m-%d_%H%M%S
+ export COMMIT=2022-12-16_141722
+ COMMIT=2022-12-16_141722
++ date +%Y-%m-%d_%H%M%S
+ export VERSION=2022-12-16_141722
+ VERSION=2022-12-16_141722
+ '[' ']'
+ '[' -z ']'
+ BUILDTOOL=(appimaged appimagetool mkappimage)
+ '[' -z amd64 ']'
+ '[' ']'
+++ dirname ./scripts/build.sh
++ cd ./scripts
++ pwd
+ PROJECT=/home/deck/build/go-appimage/scripts/..
+ '[' -z ']'
+ BUILDDIR=/home/deck/build/go-appimage/scripts/../build
+ '[' '!' -z ']'
+ mkdir -p /home/deck/build/go-appimage/scripts/../build
[...]

Issue #229, opened 2022-10-30
Installing on Fedora 36 doesn't work


hardBSDk:
I did all the installation steps but it gives permission denied errors (when the guide don't use sudo).

I tried to compile too, but "go get" needs to be "go install" and with a specified version in th end of the repo link.

CalebQ42:
Can you provide the logs for the error you're getting? When you're trying to build it yourself, make sure you're using `scripts/build.sh`.

hardBSDk:
> Can you provide the logs for the error you're getting? When you're trying to build it yourself, make sure you're using `scripts/build.sh`.

I did all the steps from the wiki, nothing left.

https://github.com/probonopd/go-appimage/blob/master/src/appimaged/README.md

PR #235, opened 2022-12-13
Desktop file FUSE mount

CalebQ42:
- Implements the idea from #112
- Desktop files and thumbnails are now stored in `~/.cache/appimaged`
- `~/.cache/appimaged/desktop` gets fuse mounted to `~/.local/share/applications/appimaged` and should unmount when appimaged stops.
  - The mount is Read-Only. If you want to edit the desktop files, you can always edit the file in .cache
  - Could make the mount writable in the future, but I couldn't get it to work  easily and doesn't seem like it's very important.
- Default to using fuse3, but falls-back to fuse2 if it's not available.


CalebQ42:
Additional fixes:

- Change build script to using stable v0.10 since the build we have been using seems to be no longer available.
- Force svg thumbnails to render at 512x512. I had an AppImage where the thumbnail was 12x12, lol.
- Fixed build instructions for the individual programs. (#230 #229)

CalebQ42:
It seems like adding PATH to the watched directory is causing erroneous desktop notifications when the system updates.
Screenshot from 2022-12-31 17-14-30
I'm going to work on the inotify code to see if I can optimize it and detect if a file is an AppImage better.

CalebQ42:
I ended up getting _a bit_ carried away. I ended up completely reworking fsnotify, integration and unintegration, and how volumes getting mounted or unmounted are handled.

- Changed to a different notify library that's actively supported.
- `appimaged` remember which AppImages it's integrated so when a file is deleted, we just have to see if it's already integrated.
  - A nice side effect is that removal notifications now have the proper application name.
  - Integration and unintegration should now be async safe by using `sync.Mutex` to make sure we are only doing one operation at a time.
- Only files that end with `.appimage` or `.app` and watched directories are checked.
- If a directory we are watching is deleted, all AppImages that were within are unintegrated.
  - If the directory is then re-created, they are re-added.
  - These give a grouped notification instead of notifications for each AppImage.
    - Ex: Added 2 applications from /mnt/Drive
- If a mount point is added or removed, we no longer re-check every single directory and re-runs integration on every AppImage, now only the effected mount point is checked or removed.
- Added `AppImage.Mount` to `goappimage`.
  - The AppImage must be type 2, and it uses fuse3.
- Application menus are only rebuilt a maximum of once a second (I think I broke this a while back).

probonopd:
> The AppImage must be type 2, and it uses fuse3.

What does this mean? This introduces a libfuse3 dependency?

CalebQ42:
It would only be for `AppImage.Mount` due to my squashfs library only using fuse3 at the moment. That being said, I had to make some changes to how my squash library handles mounts and I removed that function for now.

probonopd:
Thanks @CalebQ42. So now this PR does not introduce a FUSE3 dependency?

CalebQ42:
Nope! It prefers to use fuse3, but if that's not available it will use fuse2.

CalebQ42:
Just wondering if there's any particular reason this hasn't been pulled or if it's just that you've been busy lately. I noticed you've been working on helloSystem and it recently has a major release, so I'm guessing it's that, but just wanted to make sure. Not angry or anything, just curious.

probonopd:
Indeed @CalebQ42. Been busy, that's the main reason. Another reason is that I am a bit doubtful about the use of FUSE in general since distributions dropped libfuse2 from one day to the next when I was hoping that one could rely on libfuse2 just being there forever.

CalebQ42:
How I see it is that AppImage requires FUSE anyway so we know it's present on the system if appimaged is running at all. For that to change it would take a lot of work. Considering that both Snap and Flatpak also use FUSE and I don't think it's going anywhere anytime soon.

That being said, I'd be more then happy to separate out the bug fixes and improvements so they can be pulled in.

CalebQ42:
I'm going to separate this PR out into general fixes / improvements and FUSE

PR #243, opened 2023-01-30
Fix never-ending loop when not using kde

jsol:
When not using KDE the journal got spammed with log messages. Seems like the exit-condition for the retry loop was not set.


probonopd:
Thanks for noticing and for fixing @jsol

Issue #209, opened 2022-05-23
systemd required


vlp-git:
Hello,

I have tried to launch it on Alpine but unfortunately, it seems that systemd is required ?
Or did I miised smth to configure ?

Thanks !!

probonopd:
Yes, appimaged is supported on a certain set of Live ISOs which all use systemd, so systemd currently is a requirement.

systemd is not really central to how appimaged works, could be removed easily.

SlySven:
Is this still the case as it is a blocker for using AppImages on the **[Devuan](https://www.devuan.org/)** GNU/Linux distribution... :cry:

vadi2:
appimaged is optional to using AppImages, so if I have it right, it shouldn't be?

probonopd:
Correct, appimaged is entirely optional.

Issue #178, opened 2021-10-21
Need a way to disable all desktop notifications


ghost:
I'd prefer everything to go in a log file somewhere, I don't need desktop notifications.

Is there a way to disable these? I don't see documentation pointing me how to do this anywhere....

victoriousvictorvanquishes:
@probonopd I would need to learn Go first before submitting a patch ~ steep curve for a minor feature request!

But, I was thinking that if there existed a config option in ~/.config/systemd/user/appimaged.service ... something like  `Notifications=FALSE` that these desktop notifications would just be skipped altogether.

To me, this would seem like something an experienced Go programmer would be able to implement in a short time.

probonopd:
It's totally worth learning (and not too hard). In fact,I was learning it when I was working on go-appimage.

589290:
I too would love to see a feature like this implemented. However, I don't have any time to learn Go.

DEVLOPRR:
> It's totally worth learning (and not too hard). In fact,I was learning it "along the way" while I was working on go-appimage.

Lmao same, while working on this [project](https://github.com/DEVLOPRR/bread) I was learning go.

DEVLOPRR:
> I too would love to see a feature like this implemented. However, I don't have any time to learn Go.

Don't worry bro, I'll implement it and send a pr soon!

DEVLOPRR:
Hey @probonopd 

I've Implemented The Logic But I am confused how do i get the information that user wants to show or not show the notifications.

My Code RN:

```go
// notification.go
func sendUpdateDesktopNotification(ai *AppImage, version string, notificationInLog bool, _ string) {
	if notificationInLog {
		logWriter, err := syslog.New(syslog.LOG_SYSLOG, "AppImaged")
		if err != nil {
			log.Fatalln("Unable to set logfile:", err.Error())
		} else {
			log.SetOutput(logWriter)
			log.Println(ai.Name + " can be updated to version " + version + ".")
			return
		}
	}
        ...
}
```

CalebQ42:
> Hey @probonopd
> 
> I've Implemented The Logic But I am confused how do i get the information that user wants to show or not show the notifications.
> 
> My Code RN:
> 
> ```go
> import (
> 	...
> 	"log/syslog"
> )
> 
> // notification.go
> func sendUpdateDesktopNotification(ai *AppImage, version string, notificationInLog bool, _ string) {
> 	if notificationInLog {
[...]

PR #249, opened 2023-03-27
Reorganization

CalebQ42:
This is mostly some re-organization and cleanup loosely similar to #234 

* Moved files to be under cmd (mkappimage, appimagetool, appimaged) and pkg (goappimage) instead of everything under src.
* A bunch of cleanup suggested by linters.

And a couple features for appimaged:

* Removed notification about not running on a supported system. appimaged seems to be in a fairly stable spot and should work on most distros and it mainly seems to cause confusion.
  * Happy to re-add this is you feel it's still necessary.
* Instead of detecting if it's running in systemd move via an environment variable, it's now a separate launch command.
  * This should make it simpler if you want to use it without systemd (#209)
* Re-formatted the usage function (--help)
* Added flag to disable the mqtt service.
  * I've seen a few complaints about this, so now it can be disabled.
* Fixed the quiet flag not doing anything.
*  When installing to systemd, flags you use when launching appimaged is now added to the sytemd service.
  * Allows for the flags to be actually useful.
  * Fixes issues such as #178


probonopd:
Thanks a lot @CalebQ42. As a general rule, I'd like to avoid configurability in appimaged when possible; ideally it should "just work" for everyone in the same way.

But since so many people have reported issues with the current state of MQTT (and it is experiemental, as is the rest of the tools) I think we should disable it for now by default and only use it if explicitl requested by the user.

CalebQ42:
I think having a couple options that don't change the actual functionality could be useful. In particular, in #178 it's mentioned that disabling desktop notifications would be very nice to have. In that issue it was mentioned to check for the existence of a file (`~/.config/appimaged/no_desktop_notifications`), but a flag was already in the code so I went ahead and just hooked that up.

For the MQTT update service, I agree that it would probably be best to default it to disabled until it get some work, but it's probably best to keep it available to allow for easier testing. In these cases, would you prefer to pass-through the flags used when installing the service or simply leave it and if someone really wants to use the flags then they'll have to manually edit the service file themselves?

PR #254, opened 2023-05-28
Also look up for icons in usr/share/icons/hicolor/scalable/apps/

lephilousophe:
SVGs are traditionally stored in /usr/share/icons/hicolor/scalable/apps/ so this makes sense to look up for icons in this folder too.

### DIFF
--- a/src/appimagetool/appimagetool.go
+++ b/src/appimagetool/appimagetool.go
@@ -298,23 +298,35 @@ func GenerateAppImage(
 
 	var iconfile string
 
-	// Check if we find a png matching the Icon= key in the top-level directory of the AppDir
-	// or at usr/share/icons/hicolor/256x256/apps/ in the AppDir
+	// Check if we find a png matching the Icon= key in various locations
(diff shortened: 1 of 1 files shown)


probonopd:
Thing is, it is way easier for thumbnailers to parse/show png rather than svg. So I am not entirely sure whether we should allow svg icons. But I guess since the existing code says `supportedIconExtensions := []string{".png", ".xpm", ".svg"}` I guess I should merge this for now.

probonopd:
Maybe we should at least print a warning if if only a SVG but no PNG is found, letting the user know that a PNG should be used.

lephilousophe:
I must admit that in ScummVM we don't use a PNG on linux only a SVG icon.
By looking at [Icon Theme Specification](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html), I see that SVG is optional and XPM is deprecated so, yes, warnings may be useful.
I will add them.

lephilousophe:
And done.

probonopd:
Thanks @lephilousophe

PR #255, opened 2023-05-28
Improve ld.so.conf parsing

lephilousophe:
On Gentoo, ld.so.conf contains include directives using relative paths (as supported by ldconfig).
Add support for them.

The current implementation doesn't perfectly match what glibc does:
- glibc accepts every blank character after include, go-appimage only matches on space
- glibc split the remaining line by space or tab and includes all items, go-appimage only takes the first one and doesn't split on tabs,
- glibc ignores a hwcap directive, go-appimage takes it as a path.

This PR also fixes these.


probonopd:
Thanks @lephilousophe, great catch!

PR #259, opened 2023-07-04
Do not hardcode Qt 5

probonopd:
Maybe we can get it to work for Qt 6 applications?

Prior art:
https://github.com/probonopd/linuxdeployqt/commit/edbf092e73285d7f532082f8852b2b4b03cc399e

@CalebQ42 what do you think?

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -256,9 +256,12 @@ func AppDirDeploy(path string) {
 
 	qtVersionDetected := 0
 
-	if containsString(allELFs, "libQt5Core.so.5") == true {
-		log.Println("Detected Qt 5")
-		qtVersionDetected = 5
+	for i := 5; i <= 99; i++ {
+		if containsString(allELFs, fmt.Sprintf("libQt%dCore.so.%d", i, i)) == true {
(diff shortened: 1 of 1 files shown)


probonopd:
Download for testing:
[artifacts](https://github.com/probonopd/go-appimage/suites/14058685897/artifacts/785575955)

CalebQ42:
Looks like it should work. The only thing I would say is change for for loop checking for qt version to something like this to prevent iterating over the list potentially 94 times.

```go
for _, e := range allELFs{
  if isQt, _ := path.Match("libQt*Core.so.*", e); isQt{
    qtVersionDetected, err = strconv.Atoi(path.Ext(e))
    if err != nil{
      log.Fatal(err)
    }
  }
}
```

CalebQ42:
I haven't done much with `appimagetool`, so it's mostly just a cursory look without much knowledge of the exact workings of it. My for loop does have the potential issue of also detecting other QT libraries instead of the Core library, but it _should_ still get the correct version. If we don't _need_ to check for the core library first we could also potentially look for just `"libQt*.so.*"` to find any Qt library.

probonopd:
Tried it on a Qt 6 application and in principle it works.

PR #271, opened 2024-02-13
Make error messages in update information less vague.

Steve-Tech:
A message like "\<app\> is not a proper AppImage: Too short" kinda just isn't very helpful, so I've changed the messages in `updateinformation.go` to more helpful messages such as "Update information isn't valid".

For example this is the old message:
image
https://github.com/Ultimaker/Cura/issues/18280

Thanks for your work!
Steve


probonopd:
Thanks @Steve-Tech

PR #274, opened 2024-03-16
fix(appimaged): avoid 'Invalid desktop file' error on Debian

### LINKED ISSUE #273: 'Invalid desktop file' with exit status 1 on Debian
Hello,

I'm opening this new issue, as I feel #145 is not related, even if the error raised is the same.

I only use two AppImage apps ([czkawka](https://github.com/qarmin/czkawka) and [MuseScore](https://github.com/musescore/MuseScore)), but I feel the error comes from the way appimaged is building the desktop files and not the apps themselves.

Here is the feedback from `desktop-file-validate` (I only trimmed user-specific and app-specific elements from the output):
```
/home/<user>/.local/share/applications/appimagekit_<ref>.desktop: error: value "mv "/home/<user>/Applications/<app_name>.AppImage" ~/.local/share/Trash" for key "Exec" in group "Desktop Action Trash" contains a reserved character '~' outside of a quote
```

This error appears on a fresh Debian 12.5.0 install with XFCE desktop.
I'm also managing similar installs on Ubuntu-based OS, including Xubuntu and this error is not raised for the exact same AppImage files.

Cobraeti:
This change should fix  #273

### DIFF
--- a/src/appimaged/desktop.go
+++ b/src/appimaged/desktop.go
@@ -133,7 +133,7 @@ func writeDesktopFile(ai AppImage) error {
 			cfg.Section("Desktop Action Trash").Key("Exec").SetValue("kioclient move \"" + ai.Path + "\" trash:/")
 		} else {
 			// Provide a fallback shell command to prevent parser errors on other desktops
-			cfg.Section("Desktop Action Trash").Key("Exec").SetValue("mv \"" + ai.Path + "\" ~/.local/share/Trash/")
+			cfg.Section("Desktop Action Trash").Key("Exec").SetValue("mv \"" + ai.Path + "\" " + os.UserHomeDir() + "/.local/share/Trash/")
 		}
 
 		// Add OpenPortableHome action


CalebQ42:
It's better to use `filepath.Join` instead of string concatenation to make sure there aren't any double slashes.

Cobraeti:
I'm out this afternoon, I will try to make this rework tomorrow, thanks for the head-up 😉

Cobraeti:
Hi @CalebQ42, I just pushed the rework and also added surrounding quotes in case `os.UserHomeDir` could contain spaces (even if hope this won't happen anywhere 😅)

Cobraeti:
I will check why the pipeline failed, but that might take a while, as I don't have a Go environnement setup ATM 😅

Cobraeti:
After reading the pipeline error a few more times I think I figured the error => I passed the already concacenated string instead of a list of strings...
I hope that's enough to build, else I will really need to setup this Go environnement and check all build steps on my side to avoid bothering you 😅

CalebQ42:
I've been a bit busy so I didn't catch it initially, but `os.UserHomeDir()` returns two values so you can't inline it. Instead you'll need to set the values to a variable before you can use it.

Cobraeti:
Hi @CalebQ42, I took the time to setup Goland and it helped me even noticed that a variable `home` catching the good part of `os.UserHomeDir()` was already defined in the main package:
https://github.com/probonopd/go-appimage/blob/09fd0186774aefa2351c42b4bb22f92ce0c4f235/src/appimaged/appimaged.go#L76
I'm not sure if it's a good practice though, but I was able to compile and push it on my target system to confirm the error is now gone...

probonopd:
Thank you very much @Cobraeti

Issue #287, opened 2024-05-13
Stack overflow gathering libs


Simran-B:
appimagetool version 185

I tried to create an AppImage from a Qt 5 application built under Arch (latest Docker image) but it failed after running for almost an hour:

```
LD_LIBRARY_PATH=/root/oiio/dist/lib /root/appimagetool-831-x86_64.AppImage deploy appdir/usr/
share/applications/org.olivevideoeditor.Olive.desktop --appimage-extract-and-run
2024/05/13 09:12:31 Running inside Docker. Please make sure that the environment variables from Travis CI
2024/05/13 09:12:31 available inside Docker if you are running on Travis CI.
2024/05/13 09:12:31 This can be achieved by using something along the lines of 'docker run --env-file <(env)'.
2024/05/13 09:12:31 Please see https://github.com/docker/cli/issues/2210.
AppDir path: appdir
Exec= key contains: olive-editor
Icon= key contains: org.olivevideoeditor.Olive
2024/05/13 09:12:31 Gathering all required libraries for the AppDir...
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc04f28e388 stack=[0xc04f28e000, 0xc06f28e000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x313589?, 0xc0f9a0?})
        runtime/panic.go:992 +0x71
runtime.newstack()
        runtime/stack.go:1101 +0x5cc
runtime.morestack()
        runtime/asm_amd64.s:547 +0x8b

goroutine 1 [running]:
reflect.Value.SetUint({0x2464e0?, 0xc028c272c0?, 0x188?}, 0x7f)
        reflect/value.go:2260 +0x109 fp=0xc04f28e398 sp=0xc04f28e390 pc=0x79c789
encoding/binary.(*decoder).value(0xc04f28e928, {0x2464e0?, 0xc028c272c0?, 0x0?})
        encoding/binary/binary.go:583 +0x61a fp=0xc04f28e508 sp=0xc04f28e398 pc=0x7a871a
encoding/binary.(*decoder).value(0xc04f28e928, {0x2478e0?, 0xc028c272c0?, 0x6f9db4?})
        encoding/binary/binary.go:545 +0xa0b fp=0xc04f28e678 sp=0xc04f28e508 pc=0x7a8b0b
encoding/binary.(*decoder).value(0xc04f28e928, {0x2ebe00?, 0xc028c272c0?, 0x40?})
[...]

PR #289, opened 2024-06-14
Skip circular deps, close elf file before opening others.

chipolux:
Fixes issue #287

Skips extracting dependencies from libs we have already seen to avoid getting trapped in circular dependencies, also ensures we close the current library after we extract it's imported libs.


probonopd:
Thanks @chipolux, a very welcome contribution 👍

Issue #282, opened 2024-04-29
appimagetool don't bundle `immodules.cache` neither set `GTK_IM_MODULE_FILE`


brunvonlope:
This function is actually incomplete for pratical purposes: https://github.com/probonopd/go-appimage/blob/426deeeea1da0929c485a306a8ee7fc525551a72/src/appimagetool/appdirtool.go#L807

It's "incomplete" because `immodules.cache` file isn't copied and `GTK_IM_MODULE_FILE` isn't set in AppRun. So, the AppImage can have problems of compatibility with certain distros, according to my tests.

In GIMP testing AppImage (the word "testing" here means that we only use the AppImage for testing right now) actually needs to manually copy the file and adjusting its paths: https://gitlab.gnome.org/GNOME/gimp/-/blob/master/build/linux/appimage/bundle-gimp-appimage.sh?ref_type=heads#L171.

Reproduction steps:
1. Use `appimagetool*.appimage deploy ...` over a GTK based app
2. Observes lack of immodules.cache

probonopd:
Thanks for letting us know @brunvonlope. Most likely we need to handle `immodules.cache` in

https://github.com/probonopd/go-appimage/blob/6866993bd2a1b8b182a5f27d9f913dbd5c7aec45/src/appimagetool/appdirtool.go#L625-L675

and export `GTK_IM_MODULE_FILE` in `AppRun`.

To anyone reading this, a pull request would be highly appreciated. Thanks!

PR #295, opened 2024-07-07
Make GTK input methods work

pbs3141:
This attempts to get input methods working for packaged apps using GTK3 and GTK4.

First of all, `appdirtool.go` didn't even handle GTK4 (it was hard-coded to 2 and 3), so GTK's loadable modules weren't even being deployed. The first commit fixes this by adding support for version 4. This resulted in an error in the step "Bundling Default theme for Gtk", since there is no default theme for GTK4, so this step was restricted to versions <= 3. Along the way I noticed "Gtk .ui files found." was running multiple times (once for each GTK version), so this was fixed too.

The above still didn't get input methods working in GTK4 apps, since GTK was looking in the wrong place for its loadable modules, as checked with `GTK_DEBUG=modules`. This was related to the `GTK_EXE_PREFIX` variable being set wrong. I couldn't find any value of this variable that would work, so I deleted it and set `GTK_PATH` instead (third commit).

GTK3 also requires `immodules.cache` to be bundled and patched (unlike GTK4, which has no such file). So code was added to do this on versions <= 3 (second commit).

I have tested the above changes with Inkscape (GTK4) and confirmed that it works. Although I have tested the GTK3-specific code too, I haven't tested it with a real GTK3 app (GIMP), so could do with some help with that (@brunvonlope).

This PR is supposed to fix https://github.com/probonopd/go-appimage/issues/282. In that issue, there were two suggestions: bundle and patch `immodules.cache`, and set `GTK_IM_MODULE_FILE`. The first suggestion is done here, while the second is not since `GTK_IM_MODULE_FILE` no longer exists in GTK4 ([doc](https://docs.gtk.org/gtk4/running.html)), and we want a solution that works for all GTK versions if possible. It might be that it does need to be set though, in which case doing it for GTK <= 3 would be harmless, since it is ignored by GTK4.

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -209,14 +209,14 @@ func AppDirDeploy(path string) {
 	// GStreamer
 	handleGStreamer(appdir)
 
-	// Gtk 3 modules/plugins
-	// If there is a .so with the name libgtk-3 inside the AppDir, then we need to
+	// Gtk modules/plugins
+	// If there is a .so with the name libgtk-* inside the AppDir, then we need to
 	// bundle Gdk modules/plugins
+	deployGtkDirectory(appdir, 4)
 	deployGtkDirectory(appdir, 3)
-
-	// Gtk 2 modules/plugins
-	// Same as above, but for Gtk 2
 	deployGtkDirectory(appdir, 2)
+
+	deployGtkUiFiles(appdir)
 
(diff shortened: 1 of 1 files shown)


probonopd:
Thank you very much @pbs3141.

brunvonlope:
Sorry, @pbs3141, I wasn't able to test in time

PR #297, opened 2024-07-10
Bundle Qt plugin `platforminputcontexts`

### LINKED ISSUE #296: Lacking support for typing characters with diacritics
In general, typing accented characters (e.g. á é í ó ú) is done by pressing an accent key followed by a character, and Qt applications tend to support this. However, Qt applications bundled in an AppImage made with `go-appimagetool` do not - pressing an accent followed by a character does not generate the correct character, but rather than unaccented one.

I believe this is because the plugins that handle this are not being bundled. See [this issue](https://gitlab.com/mbunkus/mkvtoolnix/-/issues/3577) on a Qt application that formerly had this problem. This was fixed by bundling the `platforminputcontexts` plugin, as shown in [this commit](https://gitlab.com/mbunkus/mkvtoolnix/-/commit/994594853c32eac4b3c6f3f8c46597f082e865ae), so perhaps `go-appimagetool` should make sure that is bundled as well.

lambdcalculus:
Fixes #296.

I opened the issue and only after looked at the code and saw the fix was simple :D

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -1315,6 +1315,19 @@ func handleQt(appdir helpers.AppDir, qtVersion int) {
 			}
 		}
 
+        // platform plugin context - required for special characters - included if libQt5Gui.so.5/libQt6.Gui.so.6 is included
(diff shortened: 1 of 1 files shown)


probonopd:
Thanks @lambdcalculus

Issue #300, opened 2024-08-30
Error during build "Got empty qtPrfxpath, exiting"


AlbrechtL:
Hi,

I'm trying to port my project https://github.com/AlbrechtL/welle.io/ to `go-appimage` . Unfortunately, I'm getting the errors `Got empty qtPrfxpath, exiting`

More logs
```
2024/08/30 18:30:22 libraryLocations: [/home/runner/work/welle.io/Qt/6.7.2/gcc_64/lib appdir/usr/bin /usr/lib64 /lib64 /usr/lib /lib /usr/lib/x86_64-linux-gnu/libfakeroot /usr/local/lib /usr/local/lib/x86_64-linux-gnu /lib/x86_64-linux-gnu /usr/lib/x86_64-linux-gnu /lib32 /usr/lib32 /opt/hostedtoolcache/Python/3.11.9/x64/lib /usr/lib/x86_64-linux-gnu/pulseaudio /lib/x86_64-linux-gnu/pulseaudio /lib/x86_64-linux-gnu/gconv]
2024/08/30 18:30:22 len(allELFs): 336
2024/08/30 18:30:22 Adding fontconfig symlink... (is this really the right thing to do?)
2024/08/30 18:30:22 Adding AppRun...
2024/08/30 18:30:22 Find out whether Qt is a dependency of the application to be bundled...
2024/08/30 18:30:22 Detected Qt 6
2024/08/30 18:30:22 Offset of qt_prfxpath: 4908300
2024/08/30 18:30:22 Length of value of qt_prfxpath: 0
2024/08/30 18:30:22 qt_prfxpath: 
2024/08/30 18:30:22 Could not get qt_prfxpath
2024/08/30 18:30:22 Got empty qtPrfxpath, exiting
Error: Process completed with exit code 1.
```

welle.io Github actions file: https://github.com/AlbrechtL/welle.io/blob/ed68a30a6d01e2feb73cdaf91c6a8c83bd241383/.github/workflows/linux.yml
welle.io Github actions jobs: https://github.com/AlbrechtL/welle.io/actions/runs/10637457369/job/29491361238

Do you have a hint what is going wrong?

probonopd:
It looks like `qt_prfxpath` is missing or cannot be read from your Qt installation. Where did you get your Qt from? Is the environment variable `$QTDIR` set, and where is it pointing? (If it is not set or to a wrong path, you might try to set it to the folder of your Qt install that contains /plugins, /qml,...

Related:
[...]

PR #302, opened 2024-09-03
Also use $QT_ROOT_DIR

probonopd:
Also use the environment variable `$QT_ROOT_DIR`

https://github.com/probonopd/go-appimage/issues/300#issuecomment-2323463775

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -511,7 +511,12 @@ func patchQtPrfxpath(appdir helpers.AppDir, lib string, libraryLocationsInAppDir
 	f.Seek(0, 0)
 	// Search from the beginning of the file
 	search := []byte("qt_prfxpath=")
-	offset := ScanFile(f, search) + int64(len(search))
+	prfxpathPos := ScanFile(f, search)
+	if prfxpathPos < 0 {
+		helpers.PrintError("Could not find offset for " + string(search), errors.New("no " + string(search) + " token in binary"))
+		os.Exit(1)
+	}
+	offset := prfxpathPos + int64(len(search))
 	log.Println("Offset of qt_prfxpath:", offset)
 	/*
 		What does qt_prfxpath=. actually mean on a Linux system? Where is "."?
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -1492,11 +1492,14 @@ func handleQt(appdir helpers.AppDir, qtVersion int) {
 
 func getQtPrfxpath(f *os.File, err error, qtVersion int) string {
 
-	// If the user has set $QTDIR, use that instead of the one from qt_prfxpath in the library
+	// If the user has set $QTDIR or $QT_ROOT_DIR, use that instead of the one from qt_prfxpath in the library
 	qtPrefixEnv := os.Getenv("QTDIR")
+	if qtPrefixEnv == "" {
+	    qtPrefixEnv = os.Getenv("QT_ROOT_DIR")
+	}
 	if qtPrefixEnv != "" {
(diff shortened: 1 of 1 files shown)


probonopd:
@AlbrechtL does this improve the #300 situation for you when you, in other words, does it work without explicitly setting `$QTDIR` with this build? Thanks.

AlbrechtL:
Yes, it is working! See https://github.com/AlbrechtL/welle.io/actions/runs/10688777042

AlbrechtL:
I my case the new version 843 works like the previous 841 one.
[10_Create AppImage.txt](https://github.com/user-attachments/files/16878028/10_Create.AppImage.txt)

probonopd:
Let's see whether @kevle can confirm that this improves things, too - then I think it's ready for being merged.
Thanks a lot!

kevle:
I just tested the build artifacts and it seems to abort correctly now:

```
2024/09/10 10:02:34 libPath: /toolchain/exports/x64-linux/installed/x64-linux-dynamic/lib/libQt6Core.so.6
ERROR Could not find offset for qt_prfxpath=: no qt_prfxpath= token in binary
```

In my own build I tried just skipping patching libQt6Core, and Qt manages to load the platform plugins just fine:

```
diff --git a/src/appimagetool/appdirtool.go b/src/appimagetool/appdirtool.go
index 842704d..2d35306 100644
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -511,7 +511,12 @@ func patchQtPrfxpath(appdir helpers.AppDir, lib string, libraryLocationsInAppDir
        f.Seek(0, 0)
        // Search from the beginning of the file
        search := []byte("qt_prfxpath=")
-       offset := ScanFile(f, search) + int64(len(search))
+       prfxpathPos := ScanFile(f, search)
+       if prfxpathPos < 0 {
+               helpers.PrintError("Could not find offset for " + string(search), errors.New("no " + string(search) + " token in binary"))
+               return
+       }
+       offset := prfxpathPos + int64(len(search))
        log.Println("Offset of qt_prfxpath:", offset)
        /*
                What does qt_prfxpath=. actually mean on a Linux system? Where is "."?
```

In my environment, Qt is being built by vcpkg. Judging from what I understand from the design of go-appimage, it probably is out of scope to support libraries being built this way, since it expects to use system libraries.

If this is true, then aborting is probably the right thing to do to avoid scope creep.
If not, then checking for `qt_prfxpath` and only patching it when it was found is probably the way to go.

probonopd:
@kevle could you send a pull request that only does the patching if `qt_prfxpath` exists and is set? Would highly appreciate it. Thank you very much for your investigation!

kevle:
@probonopd I'll give it a go!

Samueru-sama:
> Yes, it is working! See https://github.com/AlbrechtL/welle.io/actions/runs/10688777042

I've noticed something weird, I checked the rpath of that apimage you linked and I get this: 

```
patchelf --print-rpath ./welle-io                                                                                                                                                                                                                                                                                                                                                                                        
$ORIGIN/../../home/runner/work/welle.io/Qt/6.7.2/gcc_64/lib:$ORIGIN/.:$ORIGIN/../lib64:$ORIGIN/../../lib64:$ORIGIN/../lib
...
```
That first result is wrong, it is adding a fullpath to the rpath.

probonopd:
Thanks @kevle. There is now a build for testing in #306, and it would be great if someone could give it a through test. Thank you!

kevle:
I think this should be merged. 
#306 can still be applied on top when we manage to work out a more reliable detection mechanism.

probonopd:
Thanks @kevle

PR #306, opened 2024-09-16
Support relocatable builds of Qt{5,6}

kevle:
This PR resolves a problem that leads to corrupted Qt core library files, by detecting relocatable builds of Qt and only conditionally patching the shared library files.

See https://github.com/probonopd/go-appimage/pull/302#issuecomment-2340254664


probonopd:
Thank you very much @kevle.

probonopd:
Would be great if someone could test whether this build fixes the issue. Thank you very much!

src/appimagetool/appdirtool.go
```diff
@@ -1508,21 +1533,44 @@ func getQtPrfxpath(f *os.File, err error, qtVersion int) string {
...
+
+	qt_prfxpath := ""
+	qtPrefixPathRequiresPatch := true
+
+	// When length is 0, Qt has been built as relocatable
+	if length == 0 {
+		qtPrefixPathRequiresPatch = false
+		// Directory should be in ../Qt${qtVersion}
+		// TODO: Check if this is true for relocatable binaries in Qt5
+		qt_prfxpath = filepath.Dir(library) + fmt.Sprintf("/../Qt%d", qtVersion)
```
ArcticLampyrid, line 1545: Not this case on Arch Linux.
I have `/usr/lib/libQt6Core.so` but directory on `/usr/lib/qt6`. And `qt_prfxpath` is really empty.

ArcticLampyrid: Note: Before applying this patch, it cannot work on Arch Linux either. So it is not a regression.

ArcticLampyrid: Okay, this may not be a problem because Arch's Qt is not relocatable, even though it doesn't have `qt_prfxpath`.

kevle: Thanks for testing it, I'll see if I can fix it.

kevle:
From the looks of it, there is no way around loading libQtCore6.so and querying its configuration using QLibraryInfo::location ( Qt5 ) and QLibraryInfo::path ( Qt6 ) to properly handle all cases.

I have not been able to find a reliable way yet to extract this meta information ( e.g. from embedded resource files ) any other way.

PR #307, opened 2024-09-21
Install notifications

CalebQ42:
- Added notifications when installing appimaged
  - Only shown when run without a terminal
  - Previously there was no feedback when appimaged was launched via GUI
  - Shows a fail or success message
- Added app name to notifications
  - Gnome a major release or two started always showing app name in notifications, causing all appimaged notifications to say "unknown app"
- Update deps and Go version


probonopd:
Thanks @CalebQ42. When appimaged is launched by systemd on each startup, there should _not_ be a notification. Is this the case?

CalebQ42:
Nope, it only shows the notification if it creates the service file.

probonopd:
Thanks @CalebQ42

PR #316, opened 2024-11-29
Only add directories to libraryLocationsInAppDir that contain .so files

### LINKED ISSUE #314: go-appimage treats executables as libraries to set RPATH
Running objdump -c on binaries from an .appimage produced by go-appimage reveals that the tool uses executables paths as suitables to be appended on RPATH, not only the paths of libraries.

As far I know, it is not common dynamically linkining to executables on Linux. That was a thing on very early versions of Windows but not anymore. This seems a error. 

@Samueru-sama probably can share some words too

probonopd:
Closes #314

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -11,6 +11,8 @@ import (
 	"path"
 	"strconv"
 	"syscall"
+	"regexp"
+	"io/ioutil"
 
 	"debug/elf"
 	"os"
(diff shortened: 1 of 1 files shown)


brunvonlope:
Downloading the artifact. For some reason download speed is at 80Kbps

probonopd:
Download is much faster here, maybe cancel and try again?

probonopd:
GitHub only gives artefacts downloads to logged in users. Hence I uploaded it here. Try this:

[appimagetool-856-x86_64.zip](https://github.com/user-attachments/files/17959360/appimagetool-856-x86_64.zip)

brunvonlope:
It fails with the following error:

```
ERROR ReadDir: open /builds/GNOME/gimp/AppDir/usr/lib/llvm-14/lib: no such file or directory
```

Full log: https://gitlab.gnome.org/GNOME/gimp/-/jobs/4528193

Sorry for the huge delay, our infra is in a transition phase.

brunvonlope:
According to the logs (https://gitlab.gnome.org/GNOME/gimp/-/jobs/4528886), only `lib/arch_triplet` and `lib64` are inserted in RPATH, so the situation gets worse with the patch because it can't find any other dir.

probonopd:
Argh. I need a local test case... Sorry for the noise!

PR #324, opened 2025-01-09
Fix #323

### LINKED ISSUE #323: AppImages seem to be integrated but don't show up on my KDE Application Launcher
Thanks for working on this great utility!

I'm on a Ubuntu 24.40 derivative (KDE) x86_64 and followed the directions on this github to install go-appimage daemon (cut and pasted script from github), and the daemon seemed to install fine. I then tried downloading two app images into my Downloads folder, and got a notification that the daemon detected both, but I don't see either listed in my Application Launcher. I rebooted my system a couple of times but still don't see the appimages in my Application Launcher. When I then go to move the appimages I downloaded from my Downloads folder to the trash, the appimmage daemon seems to know I did that and sends a message that they were deleted which would seem to indicate that the daemon has integrated the app images yet still they don't show up in my Application Launcher which is the main feature I was looking for from the appimage daemon.

Any suggestions?

Thanks!

CalebQ42:
* Should fix #323 due to an unhandled error.
* Changed filepaths in function to use `filepath.Join` for correctness

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/probonopd/go-appimage
 go 1.23
 
 require (
-	github.com/CalebQ42/squashfs v1.0.2
+	github.com/CalebQ42/squashfs v1.0.4
 	github.com/acobaugh/osrelease v0.1.0
 	github.com/adrg/xdg v0.5.0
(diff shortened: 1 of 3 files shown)


probonopd:
Thank you very much @CalebQ42

PR #333, opened 2025-02-24
Update README.md

yupipi93:
fix execute in zsh


probonopd:
Interesting; would we have to do this for appimagetool too?

CalebQ42:
I use zsh and the original version does not have issues on my system. @yupipi93 what error are you getting that requires this on your system?

Issue #341, opened 2025-04-23
appimaged: AppImages created using uruntime with dwarfs do not work.


CalebQ42:
The [uruntime](github.com/VHSgunzo/uruntime) to create AppImages has an option to use dwarfs archives instead of squashfs. As appimaged directly reads the squashfs archive. We can probably add a work around, but since the type 2 spec is to use squashfs, should we add in a workaround for AppImages that are saying they are a type 2 AppImage, but are not adhering to the type 2 specifications? [Issue on uruntime](https://github.com/VHSgunzo/uruntime/issues/11).

probonopd:
Well. Is there even a dwarfs filesystem for golang that we could use?

CalebQ42:
> Well. Is there even a dwarfs filesystem for golang that we could use?

No, there is no go library for dwarfs, we'd have to rely on commands.

xplshn:
> Well. Is there even a dwarfs filesystem for golang that we could use?

I could make some bindings, I've been wanting to convert the AppBundle runtime into one that doesn't have binary dependencies, at least to test how it goes and see if that improves anything. As I will still most likely keep using binaries, for ease of development & use.

Do tell me guys if y'all would be interersted in having dwarfs bindings using CGO

probonopd:
Well, so far the draft of the AppImageSpec does not specify any particular filesystem format for type-2 AppImages, but the plan has been since a while to standardize on squashfs with zstandard compression:

* https://github.com/AppImage/AppImageSpec/issues/34

However, due to the impressive results of dwarfs there has also been a consideration to standardize on it instead.

As of now, no decision has been made.

PR #342, opened 2025-04-24
Added generic type 2 appimage integration

CalebQ42:
* Added the ability to integrate non-squashfs AppImages by using the runtime arguments `--appimage-extract` and `--appimage-mount`
* Update squashfs library for increased performance
* Slightly tweaked `AppImage.Icon()` so that it will, as a last resort, return a top level png or svg if the "correct" icon can't be found.


CalebQ42:
Fixes #341. It's not an ideal solution, but should work until we have a golang dwarfs library.

probonopd:
Welll. That would introduce a huge security risk: This would execute code without the user intentionally launching the AppImage. Imagine a malicious executable masquerading as an AppImage. It could wipe ~. By just being present in the system, because appimaged would execute it.

So, I don't think we should merge this.

xplshn:
But appimaged could check for the presence of Dwarfs' magic bytes, and use `dwarfs` (the command) to mount the App to a suitable place and access its files.

CalebQ42:
> But appimaged could check for the presence of Dwarfs' magic bytes, and use `dwarfs` (the command) to mount the App to a suitable place and access its files.

That is true and will probably how it will work in the future (or at least until a go dwarfs library exits), but one of the reasons I wanted to do a more generic implementation is that, until the AppImage spec is solidified, a type 2 AppImage could _theoretically_ have any type of filesystem archive, not just limited to squashfs and dwarfs. Not to mention it was just kind of a bootstrappy, quick thing to make the dwarfs AppImage on my system to integrate properly.

xplshn:
> > But appimaged could check for the presence of Dwarfs' magic bytes, and use `dwarfs` (the command) to mount the App to a suitable place and access its files.
> 
> That is true and will probably how it will work in the future (or at least until a go dwarfs library exits), but one of the reasons I wanted to do a more generic implementation is that, until the AppImage spec is solidified, a type 2 AppImage could _theoretically_ have any type of filesystem archive, not just limited to squashfs and dwarfs. Not to mention it was just kind of a bootstrappy, quick thing to make the dwarfs AppImage on my system to integrate properly.

I'm writing a spec for AppBundles, one of the things a valid AppBundle requires is an elf section named .runtime_info, I think AppImages could copy this. Hopefully it could even be compatible and so software that integrates portable formats, like AppBundles, AppImages, FlatImages, RunImages, etc, could use the same elf section to identify the image filesystem and other data.

image
##### NOTE: I know it says json there, but the CBOR library understands what to do, same with yaml libraries, they use `json` as compat keyword because changing big structs full of `json:""` would be tedious :)

PR #343, opened 2025-04-26
Dwarfs AppImage support

CalebQ42:
* Added dwarfs implementation to goappimage using the `dwarfs` and `dwarfsextract` commands.
  * Fuse mounts to /tmp for most uses, with it automatically unmounting after 5 seconds of inactivity
* Added `dwarfs` (and symlink for `dwarfsextract`) executable to appimaged.on amd64 & arch64 (there are no 32 bit executables)
* Updated github actions to ubuntu-24.04 due to 20.04 being depreciated.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,49 +3,50 @@ name: Build All Architectures
 on:
   workflow_dispatch:
   push:
-    branches:    
-      - 'master'
+    branches:
+      - "master"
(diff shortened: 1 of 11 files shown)


probonopd:
Hi @CalebQ42. We would want static binaries for all supported architectures of go-appimage. This probably means that we will need to build our own, like we do with the other bundled tools.

PR #354, opened 2025-06-10
Make servicemenu available with newer KDE Frameworks

LightWayUp:
[Since KDE Frameworks 5.85, servicemenus are located at `$XDG_DATA_HOME/kio/servicemenus`](https://develop.kde.org/docs/apps/dolphin/service-menus/#where-the-servicemenus-are-located).  
The appimaged servicemenu desktop file is kept in the original location, and symlinks are added in new locations. This way, users of older versions of KDE apps are still supported, and newer servicemenu locations can be easily supported (if it is changed again in the future).

### DIFF
--- a/src/appimaged/filemanager.go
+++ b/src/appimaged/filemanager.go
@@ -77,6 +77,8 @@ Name=Make executable
 %s
 </actions>	
 `
+	DesktopFileName := "appimaged.desktop"
+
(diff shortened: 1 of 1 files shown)


LightWayUp:
Just tested, it does indeed work on my system with KDE Frameworks 6.x 🎉

PR #355, opened 2025-06-12
Fix race condition between AppImage daemon and GVfs/KDE Solid UDisks2 volume monitor

LightWayUp:
This fixes [issue 48 "Cannot see volumes come and go on Kubuntu 20.04"](https://github.com/probonopd/go-appimage/issues/48).

Earlier attempt to fix the issue (#233) implemented a retry-after-timeout, with the idea that the required services are simply "occasionally too slow to come up". With the 0.5s timeout, indeed most of the time the services are ready when AppImage daemon checks for their availability.
However, that doesn't actually fix the issue — a race condition present somewhere, it just masks it.

This change addresses the race condition itself via systemd dependency and ordering directives, no arbitrary sleeps are involved.

### appimaged.service systemd unit file

- Add `Slice=background.slice` to allow systemd to deprioritize appimaged background service, as AppImage daemon does not provide critical or time-sensitive services
- New line at the end of appimaged service unit file
- Fix race condition between AppImage daemon and GVfs/KDE Solid UDisks2 volume monitor

### systemd service setup prerequisite

- Update the already-installed appimaged service unit file if it is outdated  
Use `LAUNCHED_BY_SYSTEMD` environment variable value as a quick hack to check the revision of the service unit file. This is fine because the environment variable's *value* isn't being used elsewhere anyway.

> [!TIP]
> Ideally to determine whether to update the existing service unit file, its content should be compared against the content embedded in the executing version of AppImage daemon.
> The user would still be able to alter service unit properties and behaviors to their needs via "drop in" configuration file (`appimaged.service.d/local.conf`), so there should be no reason to modify the service unit file other than an actual update from AppImage daemon itself.

### End goal to achieve

AppImage daemon wants to ensure that signals from
`org.gtk.vfs.UDisks2VolumeMonitor` and `org.kde.Solid.Device` would be sent, so
it checks for relevant names on the bus.  
The check isn't necessary for AppImage daemon to monitor the messages on D-Bus, it just would like to send a notification informing the user, in the case that such messages are likely never going to be sent (due to the services being absent).

- If the volume monitoring services are installed on the system, then for the names to definitely exist on the bus when AppImage daemon checks, AppImage daemon
needs to ***start later than the services acquiring the names***.
- If the volume monitoring services are *not* installed, AppImage daemon needs to ***be able to start without them***, to send a notification.

This can be a problem, we can't simply make appimaged service only start after the volume monitoring services that might not exist.  

#### Volume monitoring services

> [!NOTE]
> - GNOME desktop and derivatives (MATE, Cinnamon) and Xfce use **GVfs UDisks 2 volume monitor**.
>     - D-Bus service name: `org.gtk.vfs.UDisks2VolumeMonitor`
>     - systemd service: `gvfs-udisks2-volume-monitor.service`
> - KDE Plasma desktop uses **Solid device integration framework, with UDisks 2 backend**.
>     - D-Bus service name: `org.kde.Solid.Device`  
>     Which doesn't exist...? So we choose the closest-related one: `org.kde.Solid.PowerManagement`, from **PowerDevil**
>     - systemd service of PowerDevil: `plasma-powerdevil.service`

Both `gvfs-udisks2-volume-monitor.service` and `plasma-powerdevil.service` are "static" services. Static services are ones with no `[Install]` section, they're only started if other units depend on them.

Take PowerDevil as example, it is depended on by `plasma-workspace.target`. `plasma-workspace-{x11/wayland}.target` are in turn also static, and only requested to be started by `startplasma-{x11/wayland}` executable binaries, which is executed by SDDM (the display manager responsible for login screen) when the user chooses a session and logs into it.  
**This means `plasma-powerdevil.service` is always only started after login.**

#### systemd "user lingering"

systemd *user* services are managed by systemd *user service manager*.

- Normally, systemd user services are run when a login session exists for that user.  
Usually this means, the user services are started after login, and stopped before logout/shutdown.
- If "user lingering" is enabled for a user, that user's user service manager is *spawned as soon as possible after boot, before any login sessions are created*.  
***The user services are started at boot***, and stopped at shutdown.  
This can be useful for long-running user services, for example [rootless Docker](https://docs.docker.com/engine/security/rootless/#daemon).

### Issue

Previously, appimaged service unit uses `WantedBy=default.target`, as in, `default.target` wants/depends on `appimaged.service` and will cause it to start.  
**`default.target` of user units is started when the user service manager is started.**

> [!NOTE]
> Not to be confused with `default.target` of system units, which is started on boot, and is often aliased to `multi-user.target` or `graphical.target` system units.

From the above, we can understand that the following sequence of events are possible:

**1.** Without user lingering:

**1. a.** AppImage daemon runs `monitorUdisks()` **before** PowerDevil acquires the name on the bus. AppImage daemon assumes that volumes can not be monitored, and sends a notification "Cannot see volumes come and go".

**1.b.** AppImage daemon runs `monitorUdisks()` **after** PowerDevil acquires
the name on the bus, everything is fine.

**2.** With user lingering:

AppImage daemon always runs `monitorUdisks()` before PowerDevil is even
started, assumes that volumes can not be monitored, and sends a notification
"Cannot see volumes come and go".

Without user lingering, a race condition exists between appimaged service and volume monitoring service.  
With user lingering, the issue is much more apparent, because appimaged service is always started before volume monitoring service.

### Solution

To solve the problem, `gvfs-udisks2-volume-monitor.service` and
`plasma-powerdevil.service` are added to `WantedBy=` *in addition to*
`default.target`.
Instead of starting AppImage daemon and checking/waiting for the availability
of volume monitoring services, now AppImage daemon is started *because of*
those services.
As well, they're added to `After=` so that appimaged service is definitely
started after them if they succeed.
`ExecCondition=` is added to skip appimaged service being started, when volume
monitoring services aren't ready yet, but are reasonably predicted to have
started at a near future point in time.

The new steps are:

1. System boots, system service manager is started.
2. Either now or after login, the user service manager is started, and starts
`default.target` and its dependencies - including `appimaged.service`.
3.a. If at least one of `gvfs-udisks2-volume-monitor.service` and
`plasma-powerdevil.service` exists and is enabled, but none are active, skip
starting `appimaged.service` now.
3.b. Else, AppImage daemon starts up, runs `monitorUdisks()`, and either
3.b.(a) enabled && active: succeeds, or
3.b.(b) !enabled: rightfully determines that volumes can not be monitored, at
least in the current implementation of AppImage daemon, and sends the
notification "Cannot see volumes come and go".
4. At the login screen, the user chooses a session and enters the password to
log in.
5. The binary set as the value of the session's `Exec=` key (for example
`startplasma-wayland`) is executed, and is responsible for setting up and
starting the rest of the session. It requests the proper user service to be
started (`plasma-workspace-wayland.target`).
6. The user service manager starts that session-specific service
(`plasma-workspace-wayland.target`) and its dependencies - including
`plasma-powerdevil.service`.
7. The volume monitoring service (`plasma-powerdevil.service`) is set to want
and start before `appimaged.service`, and either starts successfully or fails.
8. AppImage daemon starts up, runs `monitorUdisks()`, and either succeeds, or
rightfully determines that volumes can not be monitored, at least in the
current implementation of AppImage daemon, and sends the notification "Cannot
see volumes come and go".

### DIFF
--- a/src/appimaged/prerequisites.go
+++ b/src/appimaged/prerequisites.go
@@ -509,21 +509,37 @@ func installServiceFileInHome() {
 	log.Println("thisai.path:", thisai.Path)
 	d1 := []byte(`[Unit]
 Description=AppImage system integration daemon
-After=syslog.target network.target
+After=syslog.target network.target gvfs-udisks2-volume-monitor.service plasma-powerdevil.service
 
 [Service]
-Type=simple
+ExecCondition=/bin/sh -c ' \
+    unit_status() { \
+        systemctl --user --quiet "is-$${1}" "$$2"; \
+    }; \
+    some_enabled=false; \
+    some_active=false; \
+    for service in "gvfs-udisks2-volume-monitor" "plasma-powerdevil"; do \
+        service="$${service}.service"; \
+        if unit_status enabled "$$service"; then \
+            some_enabled=true; \
+            if unit_status active "$$service"; then \
+                some_active=true; \
+                break; \
+            fi; \
+        fi; \
+    done; \
+    $$some_active || ! $$some_enabled; \
+'
 ExecStart=` + thisai.Path + `
-
-LimitNOFILE=65536
-
 RestartSec=3
 Restart=always
-
 Environment=LAUNCHED_BY_SYSTEMD=1
+Slice=background.slice
+LimitNOFILE=65536
 
 [Install]
-WantedBy=default.target`)
+WantedBy=default.target gvfs-udisks2-volume-monitor.service plasma-powerdevil.service
+`)
 	err = syncWriteFile(pathToServiceDir+"appimaged.service", d1, 0644)
 	helpers.LogError("Error writing service file", err)
 
--- a/src/appimaged/prerequisites.go
+++ b/src/appimaged/prerequisites.go
@@ -405,6 +405,38 @@ func setupToRunThroughSystemd() {
 			os.Exit(0)
 		}
 
+	} else {
+		// Quick hack to check systemd service unit file revision
+		// The environment variable value isn't used anywhere else anyway
+		// Ideally, existing service unit file content should be compared
+		// against the content we would have written
+		if os.Getenv("LAUNCHED_BY_SYSTEMD") != "2" {
+			installServiceFileInHome()
+			// Re-enabling is required if [Install] section of service unit file changes
+			prc = exec.Command("systemctl", "--user", "reenable", "appimaged")
+			_, err = prc.CombinedOutput()
+			if err != nil {
+				log.Println(prc.String())
+				log.Println(err)
+				if !isatty.IsTerminal(os.Stdin.Fd()) {
+					sendDesktopNotification("Update Failed", "Unable to enable systemd service", -1)
+				}
+				os.Exit(0)
+			}
+			log.Println("Starting systemd service...")
+			prc = exec.Command("systemctl", "--user", "restart", "appimaged")
+			_, err = prc.CombinedOutput()
+			if err != nil {
+				log.Println(prc.String())
+				log.Println(err)
+				// The new version of service unit will be in effect on next boot anyway
+			} else {
+				// The current process was started by systemd as a service,
+				// and systemctl command is issued to restart the service.
+				// Is it even possible for this process to still be alive?
+				os.Exit(0)
+			}
+		}
 	}
 
 }
@@ -533,7 +565,7 @@ ExecCondition=/bin/sh -c ' \
 ExecStart=` + thisai.Path + `
 RestartSec=3
 Restart=always
-Environment=LAUNCHED_BY_SYSTEMD=1
+Environment=LAUNCHED_BY_SYSTEMD=2
 Slice=background.slice
 LimitNOFILE=65536
 


probonopd:
Thanks for the detailed analysis @LightWayUp. Would be good to hear from users on different systems how well this works.

LightWayUp:
> Thanks for the detailed analysis @LightWayUp. Would be good to hear from users on different systems how well this works.

After more investigation, I've found the root cause of the issue, that leads to the discrepancy in behavior between sending D-Bus calls to KDE Plasma's `plasma-powerdevil.service` and to GNOME's `gvfs-udisks2-volume-monitor.service`.  
I've significantly modified the original post to correct some points and add more context on the issue. In short:

- PowerDevil is supposed to be D-Bus activatable, but isn't due to a missing D-Bus service file, this is why calling its method fails for AppImage daemon
- The issue can only be fixed by KDE

The solution didn't change much though, most important part was adding a reverse dependency on `graphical-session.target` as a fallback.

---

I have tested the fix across various distributions and desktop environments, and can verify that it works correctly in all except one case, as described in the **Edge case** section above. It requires a very specific and uncommon setup to be hit, so in my opinion fixing issue #48 outweighs the potential disadvantage.

PR #360, opened 2025-07-08
deploy qt `tls` plugins

getchoo:
The `tls` plugins (which I think are just renamed from `bearer` in newer versions of Qt...?) are usually required with Qt Network. Goes well with #359 in getting more Qt app bundles working OOTB


uncor3:
just ran into this in Qt6, this should get merged imo 
```shell
qt.network.ssl: No functional TLS backend was found
qt.network.ssl: No TLS backend is available
```

probonopd:
Thanks @getchoo

PR #364, opened 2025-08-12
docs: appimagetool: document env var for Qt install root

AdelKS:
Hey,

I had to go through the code to find this information, so it's definitely useful to have in the docs

Thanks!

Adel


probonopd:
Thanks Adel.

Issue #147, opened 2021-04-09
Working directory is not as is expected


sjdrc:
We recently switched from linuxdeploy to go-appimage.

Our application creates/reads files from the current working directory.
i.e. we have the following directory
```
$ls /home/user/somedirectory
/home/user/somedirectory/Our.AppImage
/home/user/somedirectory/Some.configfile
```

previously under linux deploy, when running the AppImage, the config file would be loaded correctly.

Now when using the appimage build with `go-appimage`, the config file is not loaded. It appears that it is changing the working directory. I could not find a config option to change this.

PR #370, opened 2025-11-02
Add --preserve_cwd option to disable cds in AppRun

pbs3141:
Fixes issue https://github.com/probonopd/go-appimage/issues/147, and this error in the Inkscape AppImage making it unable to open relative paths:

```
$ ./Inkscape.AppImage myfile.svg
Can't open file: /tmp/.mount_InkscaanEjEO/myfile.svg (doesn't exist)
```

### DIFF
--- a/src/appimagetool/appdirtool.go
+++ b/src/appimagetool/appdirtool.go
@@ -38,7 +38,8 @@ var seenDeps []string
 
 var quirksModePatchQtPrfxPath = false
 
-var AppRunData = `#!/bin/sh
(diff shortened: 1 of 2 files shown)


brunvonlope:
This actually does not fix the bug. Apps that patch the LD interpreter into binaries will set --preserve_cwd to false and continue to not get the cwd 😅

pbs3141:
`--preserve_cwd` defaults to `false`, i.e. the behaviour change is entirely opt-in, so those applications will continue to work.

PR #397, opened 2026-06-05
appimage: skip .desktop with NoDisplay=true

OnkelTem:
- This pull request was created with AI assistance and should be considered a _quick fix_.
- The author is a developer, but this codebase is outside his primary expertises.

## Problem

Some AppImages ship multiple `.desktop` files (see f.e. https://github.com/probonopd/go-appimage/issues/394) but **appimaged** takes the first. In the case of Audacity **appimaged** picked the wrong one - an url handler.

## Solution

Parse .desktop files and filter out those with `NoDisplay=true` - they're barely apps one wants in the menu.


probonopd:
@OnkelTem can you please confirm that the build for testing (liked above) works for you? Thanks.

